PR #25, opened 2020-06-22
Feature/16a replace eck bundle

andybroomfield:
Adds a default bundle to the created entity.
This should fix the install.

Note: you'll have to use either devel entup and install the new config, or do a fresh install to get this working.


andybroomfield:
Thanks @danchamp am merging this one into your branch.

Issue #17, opened 2020-06-15
Replace flag dependency?


andybroomfield:
This module also relies on flag for two options, to put the banner live (publish) and to remove the hide link (it was a request from public health that Covid info on BHCC could not be hidden).

These could be replaced with simple status booleans to avoid the dependancy on flag
Flag also allows the module to use event subcriber to trigger options when put live. Right now this is to unflag any other banners once it is put live, it could also be used to trigger an email alert once a banner is put live.

andybroomfield:
One feature that flag does provide is the modal to confirm to switch on the banner, which is a feature that would be good to keep.

ekes:
The flag event also unpublishes any other alert - so there is only ever one. All this can be hung on the status, on the save. The token is also generated, but this is also done on the save by hooking in. Both can be handled in the custom entity itself.

Regarding UX. I feels like the status doesn't want to be on the edit form, at the moment in manual testing at least, I don't get the confirmation with the flag on the form. But I get the sense that a two step operation. Create, and publish separately makes sense for this operation.

So working idea:-
 * create/edit form redirects to the list of alerts
 * list of alerts shows which -if any- is currently active and each alert has link to set it (in)active
 * on clicking set alert active there is an intervening confirmation form 'Are you sure ...'

andybroomfield:
@ekes You'll get the pop up in #49.

andybroomfield:
User stories are this comment
https://github.com/localgovdrupal/localgov/issues/67#issuecomment-642956764

ekes:
> @ekes You'll get the pop up in #49.

Yer but I can (could as I've removed it ;) put the alert live with a checkbox on the form with the flag - and it doesn't do the pop-up that way - just from the flag link in the view. The user stories don't cover the banners list? 

For the story it would sound more like on the submit form, you'd want to go to a preview version (like comments do), and then confirm publish then?

andybroomfield:
@ekes Yes I think flag is only configured to do that on the view. 
I think we will still need to publish checkbox, its just it should offer a caution warning if possible. Maybe not for now as your on the banner your editing.

The admin view came later in the development process, and then get canned at BHCC for just edit one banner. However it needs the caution as you can put the banner live on that page without going into the edit page.

This is what was demoed to product group on Slack. (and discussion in #15).
[...]

Issue #16, opened 2020-06-15
Replace ECK with custom entity


andybroomfield:
As nice as ECK is we really should ship with our own entity.
We can use Drupal console to scaffold.

We need to decide if :-

1. Use a custom content entity, so as now but just not ECK.
2. We want to just go back to using state (1 banner support only)
3. Use a custom node content type, so it appears under create content.

andybroomfield:
I would favour 1 as it fits with the idea of content being stored in entities, however its not a page so it doesn't really belong on a node entity.

ekes:
Functionality is like a block, but constrained to a specific use case right?
It's even simpler than for example Status updates?

There's no access control, no page view, path? It's not such that it should feel like adding content into the site for an administrator? It might even be that it should be an 'unusual' or 'different' task?

It tend toward 1 in that case; 
although using a node content type, as status does is also an option.

ekes:
Addition: I agree it would be good to get rid of the unnecessary dependencies (eck, flag).

andybroomfield:
@ekes yes no page view, although the ECK path is currently purposed for preview purposes.
Otherwise its essentially a quick custom block for displaying persisntatly accross the top of the page. Its currently reachable on install under content -> alert banner.

andybroomfield:
Also should rename the entity from alert_banner to `localgov_alert_banner`

andybroomfield:
See also #18. Fields need to be renamed.

andybroomfield:
There is also the admin view to update at http://localgov.lndo.site:8000/admin/content/alert-banners and the display view http://localgov.lndo.site:8000/admin/structure/views/view/alert_banner
The flag `put_live` will need to updated to the new entity.
The template will also need to update `eck-entity--alert-banner.html.twig`

andybroomfield:
For nameing conventions see https://github.com/localgovdrupal/localgov/wiki/Naming-conventions

danchamp:
That all makes sense. Presumbly also move remove_hide_link to the entity to remove the flag dependency?

andybroomfield:
@danchamp Yes, think the hide link can go. I'd like to keep put live as a flag for now to keep the modal confirm behaviour until we tackle #17.

andybroomfield:
Fixed with #26. Other issues added as followup.

PR #26, opened 2020-06-22
Feature/16 replace eck

### LINKED ISSUE #18: Rename fields to use the localgov format
Follow the example in localgov_services and rename `field_` to `localgov_alert_`...
(Check format before adding).

danchamp:
- Removes dependency on ECK #16 and supports multiple banner records (one live)
- Replaces flag with field for hide link #17 
- Ref #18


andybroomfield:
Thanks for working on this @danchamp.
This is good for the first change. Its opened up a fair few issues, however they are better tackled seperatly.
I've fixed the test to work with the renamed entities.
I need to review with @ekes and @stephen-cox the field names they used on services and if they prefixed them, as these ones arn't which is why I altered your comment to stop the issue auto closing. We can tackle that seperatly too, however these arn't changeable fields so it should actully be ok to leave them unprefixed.

danchamp:
> We need to remove eck as a dependancy from `localgov_alert_banner.info.yml` and `composer.json`

Thanks @andybroomfield that's those changes made.

PR #33, opened 2020-06-23
Fix code standards accross alert banner

### LINKED ISSUE #5: Run and fix code standards issues

### LINKED ISSUE #24: Use dependancy injection instead of \drupal calls
Follow up to #6.
Although this ignored in the phpstan config ignores these calls, it something we should fix, though after the general refactor in #16 is done.

Initial report of running phpstan 
````
 ------ ---------------------------------------------------------------------- 
  Line   src/Controller/AlertBannerController.php                              
 ------ ---------------------------------------------------------------------- 
  37     \Drupal calls should be avoided in classes, use dependency injection  
         instead                                                               
 ------ ---------------------------------------------------------------------- 

 ------ ---------------------------------------------------------------------- 
  Line   src/EventSubscriber/AlertBannerFlagLive.php                           
 ------ ---------------------------------------------------------------------- 
  31     \Drupal calls should be avoided in classes, use dependency injection  
         instead                                                               
  40     \Drupal calls should be avoided in classes, use dependency injection  
         instead                                                               
  53     \Drupal calls should be avoided in classes, use dependency injection  
         instead                                                               
  69     \Drupal calls should be avoided in classes, use dependency injection  
         instead                                                               
  75     \Drupal calls should be avoided in classes, use dependency injection  
         instead                                                               
 ------ ---------------------------------------------------------------------- 

                                                                                
 [ERROR] Found 6 errors          
````

andybroomfield:
Fix #5
Fix #24

Auto applied fixes with lando phix.
Manually added documentation comments in format as required
Replace class \Drupal calls with dependancy injection

### DIFF
--- a/localgov_alert_banner.install
+++ b/localgov_alert_banner.install
@@ -1,0 +1,33 @@
+<?php
+
+/**
+ * @file
+ * Contains localgov_alert_banner.install.
+ */
+
+/**
+ * Implements hook_uninstall().
+ */
+function localgov_alert_banner_uninstall() {
+
+  // Remove configuration.
+  $configs = [
+    "block.block.views_block__alert_banner_alert_banner_block",
+    "core.entity_form_display.alert_banner.alert_banner.default",
+    "core.entity_view_display.alert_banner.alert_banner.default",
+    "flag.flag.set_live",
+    "system.action.flag_action.set_live_flag",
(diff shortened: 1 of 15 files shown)


andybroomfield:
@stephen-cox looks like that deprecations introduced by drupal console. Have push amended version.

Issue #35, opened 2020-06-24
Restore the separate permission to view all alert banners


andybroomfield:
During the conversion, the permission to view the alert banner list was removed.
This should be restored and access to the alert banner list (or the admin view if restored) should use this permission.
The permission is `access alert banner listing page` which would have been provided by ECK.

Issue #28, opened 2020-06-22
Restore the admin view for alert banners


andybroomfield:
The new custom entity is using standard drupal entity list for showing the banners that can be edited. This replaces the custom admin view that includes features for putting live in modal.

This should be restoried in place of the drupal default (same method content uses) and linked to the alert banners menu item, but using the new entity.

Previous Yaml is
````yaml
langcode: en
status: true
dependencies:
  config:
    - eck.eck_type.alert_banner.alert_banner
    - field.storage.alert_banner.field_alert_type_of_alert
    - flag.flag.put_live
    - system.menu.admin
  module:
    - eck
    - flag
    - options
    - user
id: admin_manage_alert_banners
label: 'Manage Alert Banners'
module: views
description: ''
tag: ''
base_table: alert_banner_field_data
base_field: id
core: 8.x
display:
  default:
    display_plugin: default
    id: default
    display_title: Master
    position: 0
    display_options:
      access:
        type: perm
        options:
          perm: 'access alert_banner entity listing'
      cache:
        type: tag
        options: {  }
      query:
        type: views_query
        options:
          disable_sql_rewrite: false
          distinct: false
          replica: false
          query_comment: ''
          query_tags: {  }
      exposed_form:
        type: basic
        options:
          submit_button: Filter
          reset_button: false
          reset_button_label: Reset
          exposed_sorts_label: 'Sort by'
          expose_sort_order: true
          sort_asc_label: Asc
          sort_desc_label: Desc
      pager:
        type: mini
        options:
          items_per_page: 10
          offset: 0
          id: 0
          total_pages: null
          expose:
            items_per_page: false
            items_per_page_label: 'Items per page'
            items_per_page_options: '5, 10, 25, 50'
            items_per_page_options_all: false
            items_per_page_options_all_label: '- All -'
            offset: false
            offset_label: Offset
          tags:
            previous: ‹‹
            next: ››
      style:
        type: table
        options:
          grouping: {  }
          row_class: ''
          default_row_class: true
          override: true
          sticky: false
          caption: ''
          summary: ''
          description: ''
          columns:
            title: title
            field_alert_type_of_alert: field_alert_type_of_alert
            uid: uid
            created: created
            changed: changed
            link_flag: link_flag
            operations: operations
          info:
            title:
              sortable: true
              default_sort_order: asc
              align: ''
              separator: ''
              empty_column: false
              responsive: ''
            field_alert_type_of_alert:
              sortable: true
[...]

Issue #27, opened 2020-06-22
Fix display template for alert banner


andybroomfield:
Now its a new entity we need to fix the template that is used to the alert banner style.

Examples (BHCC)
<img width="1048" alt="Screenshot 2020-06-22 at 6 27 33 pm" src="https://user-images.githubusercontent.com/1467480/85317333-22da4980-b4b6-11ea-8dae-86400315767b.png">
Minor

<img width="1049" alt="Screenshot 2020-06-22 at 6 27 21 pm" src="https://user-images.githubusercontent.com/1467480/85317328-21a91c80-b4b6-11ea-930a-2935286651ac.png">
Major

We should only offer a basic templating in the module, as themes will want to override it.
Make sure we supply a class name for the type of alert so themes can approitally colour them.

andybroomfield:
Is missing a hook theme in the .module file.
Previously would have been
````php
/**
 * Implements hook_theme().
 */
function localgov_alert_banner_theme() {
  return [
    // Bundle our template with the module (which can be overridden in theme)
    'eck_entity__alert_banner' => [
      'base hook' => 'eck_entity',
    ],
  ];
}
````

andybroomfield:
Before we do this, we need to make the entity fieldable and restore fields to it (blocked by #29)

andybroomfield:
For reference, it is this [GDS pattern ](https://docs.publishing.service.gov.uk/manual/emergency-publishing.html#death-of-a-notable-person) that we are implementing.

In addition to what is on the GDS style guide, we have a hide button / link that should use JS to hide the banner (there is included JS in the ported module). This should *not* be displayed if the field `remove_hide_link` is checked on the edit form.

to support sites that do not show the alert banner title, there is now an additional field `display_title` like blocks, so the alert banner needs to account for this. Additionally a difference between the GDS pattern and BHCC is that, on minor alert and annoucements only, is that the title is inline.

andybroomfield:
@danchamp Try to fix #38 in the same PR and let me know if that is possible.

PR #37, opened 2020-06-25
Regenerate the alert banner entity as a fieldable entity bundle

### LINKED ISSUE #29: Allow banner types to be fieldable
The alert banners will just use the default fields at present.
Councils may want to add their own fields, so we should allow access to the manage fields and manage display ui like eck.

### LINKED ISSUE #34: Non standard table field for uid on localgov_alert_banner entity
The newly created entity for the alert banner uses a non standard field `user_id` or the user reference.
This needs to be replaced with the Drupal standard `uid`

andybroomfield:
Fix #29
Fix #34

Alert banner generated as entity localgov_alert_banner
with default bundle localgov_alert_banner

Fields namespaced into localgov
Adds new display title checkbox


andybroomfield:
@ekes @stephen-cox I'm still a little confused about field namespacing.
The fields are field_localgov_* (This seems to be the pattern I see in the services modules).

andybroomfield:
Merging this PR will unbock #27 and #28.

config/install/core.entity_form_display.localgov_alert_banner.localgov_alert_banner.default.yml
```diff
@@ -0,0 +1,76 @@
+langcode: en
+status: true
+dependencies:
+  config:
+    - field.field.localgov_alert_banner.localgov_alert_banner.field_localgov_link
```
ekes, line 5: Yes so if you compare this to https://github.com/localgovdrupal/localgov_services/blob/b8d65a1a361a05ec368f3156f7d818c12bc93fff/modules/localgov_services_page/config/install/core.entity_form_display.node.localgov_services_page.default.yml#L13

The field name `localgov_services_parent`, the bundle `localgov_services_page`, and the entity type `node`.

config/install/field.field.localgov_alert_banner.localgov_alert_banner.field_localgov_link.yml
```diff
@@ -0,0 +1,22 @@
+langcode: en
+status: true
+dependencies:
+  config:
+    - field.storage.localgov_alert_banner.field_localgov_link
```
ekes, line 5: So continuing to compare https://github.com/localgovdrupal/localgov_services/blob/b8d65a1a361a05ec368f3156f7d818c12bc93fff/modules/localgov_services_page/config/install/field.field.node.localgov_services_page.localgov_services_parent.yml#L5

The storage, which in the case of services parent is shared between multiple services node bundles so is in the localgov_services_navigation module, is name `localgov_services_parent` and entity `node`.

config/install/field.field.localgov_alert_banner.localgov_alert_banner.field_localgov_link.yml
```diff
@@ -0,0 +1,22 @@
...
+    - localgov_alert_banner.localgov_alert_banner_type.localgov_alert_banner
+  module:
+    - link
+id: localgov_alert_banner.localgov_alert_banner.field_localgov_link
+field_name: field_localgov_link
```
ekes, line 10: https://github.com/localgovdrupal/localgov_services/blob/b8d65a1a361a05ec368f3156f7d818c12bc93fff/modules/localgov_services_page/config/install/field.field.node.localgov_services_page.localgov_services_parent.yml#L12

The confusing might arise from the fact that most of the fields haven't been touched and are still `field_something`. But here it's keeping with the `localgov_`

config/install/field.storage.localgov_alert_banner.field_localgov_type_of_alert.yml
```diff
@@ -0,0 +1,32 @@
+langcode: en
+status: true
+dependencies:
+  module:
+    - localgov_alert_banner
+    - options
+id: localgov_alert_banner.field_localgov_type_of_alert
```
ekes, line 7: And so finally the storage, https://github.com/localgovdrupal/localgov_services/blob/b8d65a1a361a05ec368f3156f7d818c12bc93fff/modules/localgov_services_navigation/config/install/field.storage.node.localgov_services_parent.yml#L7

As for naming _these_ fields. I actually think you're free to call them whatever you want. Because the entity type itself is `localgov_`. I'd drop the `field_` bit just because it's not necessary, but it's not strictly required to include the `localgov_` bit.

andybroomfield: As per slack conversation, I've unprefixed the fields.
These match up with the base entity definations, and as user fields will typically be `field_` and modules their own prefix its safe to assume it won't conflict here. If users are changing or removing their own sites field prefixes, we can assume they understand the risks.

src/AlertBannerEntityAccessControlHandler.php
```diff
@@ -0,0 +1,124 @@
...
+   */
+  protected function checkAccess(EntityInterface $entity, $operation, AccountInterface $account) {
+    /** @var \Drupal\localgov_alert_banner\Entity\AlertBannerEntityInterface $entity */
+
+    switch ($operation) {
+
+      case 'view':
+
+        if (!$entity->isPublished()) {
+          $permission = $this->checkOwn($entity, 'view unpublished', $account);
```
ekes, line 28: I guess this can become another issue, but will need tests for permissions. I wonder if they need to be so complex? Can they not just be a permission for who can access unpublished ones and everyone else who can't?

andybroomfield: I think permissions are going to be a seperarte issue. 
This is more an issue of what the scaffold generates, and there might be a need for the more granulated permissions anyway, so I'm not sure its something to worry about for alpha.
In terms of the workflow, at BHCC we allow all comms to write the banner, but only the emergency publisher role to publish. Though whether we should define that as a seperate role or a new permissions is something worth investigating. We also have the permission in #35.

As this blocks a lot of issues, i'd rather move this forward and open seperate issues to trim down what we don't need. We do need to write seperate tests to make sure only the right access is granted to author, edit, delete and publish, but first we need to discuss what thos should be.

Issue #38, opened 2020-06-27
Remove display title and remove hide link from banner display.


andybroomfield:
The variables should still be available for themes, but they should never be displayed.
<img width="216" alt="Screenshot 2020-06-27 at 2 06 15 pm" src="https://user-images.githubusercontent.com/1467480/85923014-829e6f00-b87f-11ea-937d-578403f88ae4.png">

andybroomfield:
Possible by removing the following from the alert banner base fields
````php
      ->setDisplayOptions('view', [
        'label' => 'above',
        'type' => 'string',
        'weight' => -4,
      ])
````
Need to check if this keeps the variables accessible in twig for themers.

PR #40, opened 2020-06-30
Fix/27 display template

danchamp:
This pull request provides a basic styled display template for alert banners #27 , and removes non-content fields from display #38 

There are several issues that emerge from this that I'm not sure about:

- Per-alert type styling - alert type keys now include weighting, if this is stable can we just use for example '20--minor' as the class to target?
- The method of defaulting to 'More information' as the link title if none is set - currently this is done in twig, would it be better in the template preprocess?
- Extraneous markup from views when rendered - how is this removed?

### DIFF
--- a/templates/localgov_alert_banner.html.twig
+++ b/templates/localgov_alert_banner.html.twig
@@ -42,13 +42,13 @@
           <div class="localgov_alert_banner--content-link">{{ content.link }}</div>
         {% endif %}
       </div>
-      <div class="localgov_alert_banner--actions paragraph">
-        {% if not remove_hide_link %}
+      {% if not remove_hide_link %}
+      <div class="localgov_alert_banner--actions">
           <div class="localgov_alert_banner--dismiss">
             <a href="#" class="js-alert-banner-close">Hide <span class="sr-only">Notifications</span></a>
           </div>
-        {% endif %}
       </div>
+      {% endif %}
     </div>
   </div>
 </div>
--- a/templates/localgov_alert_banner.html.twig
+++ b/templates/localgov_alert_banner.html.twig
@@ -9,14 +9,46 @@
  * Available variables:
  * - content: A list of content items. Use 'content' to print all content, or
  * - attributes: HTML attributes for the container element.
+ * - display_title: Boolean indicating if alert banner title should be displayed
+ * - remove_hide_link: Boolean indicating if the link to close the banner should be hidden
+ * - type_of_alert: Type of alert banner
  *
  * @see template_preprocess_localgov_alert_banner()
  *
  * @ingroup themeable
  */
 #}
-<div{{ attributes.addClass('localgov_alert_banner') }}>
-  {% if content %}
-    {{- content -}}
-  {% endif %}
+{{ attach_library('localgov_alert_banner/alert_banner') }}
+{%
+  set has_link = content.link is not empty
+%}
+{% set classes = [
+  'js-alert-banner',
+  'localgov_alert_banner--' ~ type_of_alert,
+  is_front ? 'localgov_alert_banner--homepage' : '',
+  has_link ? 'localgov_alert_banner--has-link' : 'localgov_alert_banner--no-link'
+] %}
+<!-- Alert Banner-->
+<div class="localgov_alert_banner" role="alert">
+  <div{{ attributes.addClass(classes) }}>
+    <div class="localgov_alert_banner--inner">
+      <div class="localgov_alert_banner--inner--content">
+        {% if display_title.value %}
+          <h2 class="localgov_alert_banner--content-heading">{{ content.title }}</h2>
+        {% endif %}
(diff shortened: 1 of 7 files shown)


andybroomfield:
Quick answers, I will try to review this evening.

> * Per-alert type styling - alert type keys now include weighting, if this is stable can we just use for example '20--minor' as the class to target?
I don't think you can start classes with numbers, the keys are there to help with ordering and sorting in admin views (#28) If possible try to strip out the number, but should be fine if not (we assume if a dev is changing these, they know they will after alter the css).
> * The method of defaulting to 'More information' as the link title if none is set - currently this is done in twig, would it be better in the template preprocess?
Good point! Thats a BHCC thing as this comes from our theme, but as these are more generic template, it would be more helpful for themers to be able to use `{{ content.link }}`
> * Extraneous markup from views when rendered - how is this removed?
BHCC (and presumably Croydon & Localgov_theme) have custom markup for rending views. I'd leave that for now as how views markup should be rendered is a themes thing. As only a single banner gets presented, I've been thinking if views is a bit too much of an overhead and we could just use an entitiyquery to find the banner to present and render that in a custom block, thats another issue however.

danchamp:
Thanks @andybroomfield 

> > * Per-alert type styling - alert type keys now include weighting, if this is stable can we just use for example '20--minor' as the class to target?
> 
> [I don't think you can start classes with numbers](https://benfrain.com/when-and-where-you-can-use-numbers-in-id-and-class-names/), the keys are there to help with ordering and sorting in admin views (#28) If possible try to strip out the number, but should be fine if not (we assume if a dev is changing these, they know they will after alter the css).

Good point about valid class names. Easy enough to split the key, I'll add this.

> > * The method of defaulting to 'More information' as the link title if none is set - currently this is done in twig, would it be better in the template preprocess?
> 
> Good point! Thats a BHCC thing as this comes from our theme, but as these are more generic template, it would be more helpful for themers to be able to use `{{ content.link }}`

Thanks, that's what I thought. I'll take a look at this too.

danchamp:
- Type of alert class name now derived from keys without numeric prefix. Eg. for key `20--minor` class is `localgov-alert-banner--minor`
- Default link text 'More information' now defined in template_preprocess, so themers can use `{{ content.link }}`

templates/localgov_alert_banner.html.twig
```diff
@@ -9,14 +9,47 @@
...
+%}
+<!-- Alert Banner-->
+<div class="localgov_alert_banner" role="alert">
+  <div{{ attributes.addClass(classes) }}>
+    <div class="localgov_alert_banner--inner">
+      <div class="localgov_alert_banner--inner--content">
+        {% if display_title %}
+          <h2 class="localgov_alert_banner--content-heading">{{ content.title }}</h2>
+        {% endif %}
+        <p class="localgov_alert_banner--content-body">{{ content.short_description }}</p>
```
andybroomfield, line 40: As its now a wysiwyg element, should be a `<div>`.
Also as users might add new fields, I'd use {{ content|without(link) }}

andybroomfield: We might also consider just rending `{{ content }}` here instead of the link separately now that we are controlling the link title in a preprocesss (the special theming for the link is a very BHCC thing).

templates/localgov_alert_banner.html.twig
```diff
@@ -9,14 +9,47 @@
...
+        {% if content.link %}
+          <div class="localgov_alert_banner--content-link">{{ content.link }}</div>
+        {% endif %}
+      </div>
+      <div class="localgov_alert_banner--actions paragraph">
+        {% if not remove_hide_link %}
```
andybroomfield, line 46: Ideally move the if outside the wrapper div block so it only renders if the hide link needs to be.
We can also remove the paragraph class as that comes from BHCC theme.

danchamp: Thanks @andybroomfield that all makes sense, I'll get on with the changes.

templates/localgov_alert_banner.html.twig
danchamp: This is done, I've left the link as rendered separately for now, to make styling more straightforward in case the short_description includes links.

tests/src/Functional/AlertBannerBlockTest.php
```diff
@@ -76,4 +76,69 @@ public function testNonLiveAlertBannerDoesNotDisplay() {
...
+
+    // Flag the alert banner to put it live.
+    $flag_service = $this->container->get('flag');
+    $flag = $flag_service->getFlagById('localgov_put_live');
+    $flag_service->flag($flag, $alert);
+
+    // Check hide link is shown when remove_hide_link is not set
+    $this->drupalGet('<front>');
+    $this->assertSession()->responseContains('js-alert-banner-close');
+    $this->assertSession()->pageTextContains('Hide');
```
andybroomfield, line 135: Not sure we need this as we already test for `->responseContains('js-alert-banner-close');`

PR #44, opened 2020-07-02
Update uninstall config list

### LINKED ISSUE #30: Fix uninstall config for entity bundle
`localgov_alert_banner.localgov_alert_banner_type.localgov_alert_banner.yml` needs to be added to localgov_alert_banner_uninstall.

danchamp:
Closes #30


localgov_alert_banner.install
```diff
@@ -12,14 +12,21 @@ function localgov_alert_banner_uninstall() {
...
+    "field.field.localgov_alert_banner.localgov_alert_banner.type_of_alert.yml",
+    "field.storage.localgov_alert_banner.link.yml",
+    "field.storage.localgov_alert_banner.short_description.yml",
+    "field.storage.localgov_alert_banner.type_of_alert.yml",
+    "flag.flag.localgov_put_live.yml",
+    "localgov_alert_banner.localgov_alert_banner_type.localgov_alert_banner.yml",
+    "system.action.flag_action.localgov_put_live_flag.yml",
+    "system.action.flag_action.localgov_put_live_unflag.yml",
+    "user.role.emergency_publisher.yml",
+    "views.view.alert_banner.yml",
```
andybroomfield, line 29: These should not include the `.yml` at the end.

danchamp: Doh. Updated.

danchamp:
> Cancel that, config not being removed.
> user.role.emergency_publisher

Confirmed here, thanks @andybroomfield . The only way I can get the role to be removed on uninstall is to add the module as dependency in the role install yaml:

`dependencies:
  module:
    - localgov_alert_banner
`
I can see some other configs use this, but not all. Would adding this to all configs make the uninstall hook function redundant?

andybroomfield:
@danchamp At BHCC we use the declare module dependancy in config method, which is why I think some of these config files have it. Best to check with @ekes @stephen-cox and @finnlewis for what the LocalGov Drupal way will be, to make sure we follow the same method as the other modules.

andybroomfield:
Confirmed that these should be added in the config and we can remove this hook.
Use the following so the config gets removed during uninstall.
````yaml
dependencies:
  enforced:
    module:
      - localgov_alert_banner
````

**Comment by danchamp** [comment]
The module dependency has been added to each config. I've left the call to uninstallEntityType in the uninstall hook, is there a better way to do this?

File: config/install/block.block.views_block__alert_banner_alert_banner_block.yml [hunk 2]
```diff
@@ -4,6 +4,7 @@ dependencies:
   config:
     - views.view.alert_banner
   module:
+    - localgov_alert_banner
     - views
```
andybroomfield, line 8: I would try and replace it with.
````yaml
module:
    - views
enforced:
    module:
      - localgov_alert_banner
````
This is because, when a module is uninstalled, the config can still be left behind. This guarantees the config gets removed.

andybroomfield:
> The module dependency has been added to each config. I've left the call to uninstallEntityType in the uninstall hook, is there a better way to do this?

I think thats fine, given the entity is built in code I would expect it to be removed that way.
It's really the config uninstall that needed looking at.

config/install/block.block.views_block__alert_banner_alert_banner_block.yml
danchamp: Ah, thanks @andybroomfield , understood and updated.

andybroomfield:
I now get the following error
````
In Connection.php line 701:
  SQLSTATE[42S02]: Base table or view not found: 1146 Table 'database.localgov_alert_banner' doesn't exist: SELECT base_table.vid AS vid, base_table.id AS id  
  FROM                                                                                                                                                         
  {localgov_alert_banner} base_table                                                                                                                           
  INNER JOIN {localgov_alert_banner_field_data} localgov_alert_banner_field_data ON localgov_alert_banner_field_data.id = base_table.id                        
  WHERE localgov_alert_banner_field_data.type = :db_condition_placeholder_0; Array                                                                             
  (                                                                                                                                                            
      [:db_condition_placeholder_0] => localgov_alert_banner                                                                                                   
  )  
````
and
````
In Statement.php line 59:
SQLSTATE[42S02]: Base table or view not found: 1146 Table 'database.localgov_alert_banner' doesn't exist  
````

andybroomfield:
I've updated this PR removing the .install file. 
This seems to work in terms of also removing the entity automatically, presumably by no longer being defined or config being removed.

danchamp:
Thanks @andybroomfield

PR #46, opened 2020-07-03
Fix entity_type_update and entity_type_create hook function names

### LINKED ISSUE #45: hook_ENTITY_TYPE_update and hook_ENTITY_TYPE_create function names
Functions incorrectly named and not being called when banners updated / created:
https://github.com/localgovdrupal/localgov_alert_banner/blob/9f71319889e015b95a06df063a0a0404db677c27/localgov_alert_banner.module#L88-L100

Should be localgov_alert_banner_localgov_alert_banner_create and localgov_alert_banner_localgov_alert_banner_update

danchamp:
Closes #45

### DIFF
--- a/localgov_alert_banner.module
+++ b/localgov_alert_banner.module
@@ -86,23 +86,23 @@ function localgov_alert_banner_preprocess(&$variables) {
 }
 
 /**
- * Implements hook_alert_banner_create().
+ * Implements hook_ENTITY_TYPE_create().
  */
-function localgov_alert_banner_alert_banner_create(EntityInterface $entity) {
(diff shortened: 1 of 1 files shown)


danchamp:
The functions aren't called for me when using `localgov_alert_banner_create` and `localgov_alert_banner_update` so the token doesn't reset when updating. Using `{module_name}_{entity_type}_create` and _update they are getting called.

I haven't done anything with bundles yet - any ideas how this can be abstracted to apply to all alert banner bundles?

andybroomfield:
@danchamp That actully might be correct as its module then entity and should apply to all.
I'll create a bundle and test.

PR #48, opened 2020-07-03
Restore the separate permission to view all alert banners

### LINKED ISSUE #35: Restore the separate permission to view all alert banners

danchamp:
Closes #35


danchamp:
Thanks @andybroomfield. I think the extra menu item came from the Drupal console scaffolding.

PR #51, opened 2020-07-06
Check link field and type of alert fields exist before preprocessing

### LINKED ISSUE #50: InvalidArgumentException: Field link is unknown. if link field is removed
Error seems related to moving link in preprocess_localgov_alert_banner.
This was discovered whilst viewing a banner in a new alert type, though will also occur if the link field is removed.

The preprocess should check the link field exists first before trying to alter it.

danchamp:
Closes #50

### DIFF
--- a/localgov_alert_banner.page.inc
+++ b/localgov_alert_banner.page.inc
@@ -34,14 +34,19 @@ function template_preprocess_localgov_alert_banner(array &$variables) {
   /** @var AlertBannerEntity $entity */
   $entity = $variables['elements']['#localgov_alert_banner'];
 
-  // Set default link text
-  if ($entity->get('link')->title === '') {
-    $variables['content']['link'] = Link::fromTextAndUrl(t('More information'), Url::fromUri($entity->get('link')->uri))->toString();
+  // Set default link text if field exists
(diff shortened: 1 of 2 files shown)


andybroomfield:
I'm seeing the following when creating a new alert banner entity of a test type.
Screenshot 2020-07-07 at 9 54 07 am

danchamp:
That's expected behaviour for a new default entity type I think. What's rendered is the title field with label visible, and the admin link is the Authored by field which renders without a label by default. Presumably we'd expect these to be over-ridden in 'Manage display' for the entity type, or via twig?

andybroomfield:
> That's expected behaviour for a new default entity type I think. What's rendered is the title field with label visible, and the admin link is the Authored by field which renders without a label by default. Presumably we'd expect these to be over-ridden in 'Manage display' for the entity type, or via twig?

Yes, just checking those. @danchamp.
It clears the error so I'm good for these to be merged.

Screenshot 2020-07-07 at 10 03 35 am

How we might set up default fields and manage display can be another issue.
Screenshot 2020-07-07 at 10 03 17 am

Ideally we also could do with a test, but that can be added later.

PR #55, opened 2020-07-07
Remove flag dependency #17.

ekes:
* Relies on published status to not/display in view.
 * Moves responsibility for unpublishing other alerts to the entity save.
 * Code for save, and form moved to the custom entity and form code rather than hooks.

Follow-up still to do:

For now the 'published' checkbox is shown on the entity form.
This requires 'confirmation' and/or an action on the action admin entity list form.


ekes:
Looks like it'll need a merge already. So the sooner this commit goes in, can then continue with the confirmation step part. I'll sort the merge conflict in a moment.

ekes:
OK so that removes the field from the newly reincluded admin view too, but will be updated with the confirmation publish action.

andybroomfield:
@ekes is the removal of the column to say which banner is live intentional? I know thats where the flag was. I think we still need a column in the admin view to show which banner is currently live.

ekes:
Yes because it was working on the flag to define if it was live rather than if it's published. Need to update this view to - publish rather than flag - add confirmation - and display.

andybroomfield:
Ok, do you mind if I add an extra field for now just to display it? Then this should be good to go.
Screenshot 2020-07-07 at 10 03 35 am

ekes:
Sure makes sense if it's running on the published status.

PR #58, opened 2020-07-07
Replace hide link with button

### LINKED ISSUE #43: Accessibility: replace <a> element for hiding alert with <button>
Using an anchor element with a null (#) href to hide alert banners has accessibility implications:

- Semantically, an anchor indicates a hyperlink that takes the user somewhere else, either on the same page via a valid fragment identifier, or to another location via a valid URL.
- For keyboard users, hyperlinks once focussed are activated with the enter key, while the space bar will scroll the view down one page. Buttons are activated with either key. If the link was styled like a button this could cause user confusion.
- Links will be announced by most screen readers as 'Link' + link text, so by default 'Link hide notifications'. (The inclusion of 'notifications' for sr-only is a bit ambiguous). The hide link will also be collected by most SR software into the list of page links.

A button element will provide the same functionality as an anchor, without these accessibility issues.

danchamp:
Closes #43:

- Replaces 'hide' link with a button element
- Adds 'aria-hidden' attribute to banner on click
- Basic styling based on Croydon
- Updated tests

### DIFF
--- a/css/localgov-alert-banner.css
+++ b/css/localgov-alert-banner.css
@@ -9,13 +9,29 @@
 .localgov_alert_banner--inner {
   display: flex;
   justify-content: space-between;
-  margin: 0rem 0.9375rem;
-  padding: 0.625rem 0rem;
+  margin: 0 0.9375rem;
+  padding: 0.625rem 0;
 }
 .localgov_alert_banner--actions {
   display: flex;
   justify-content: space-between;
   align-items: flex-end;
+}
+.localgov_alert_banner--actions button {
+  background: #77bb55;
+  border: none;
+  color: #000;
+  font-weight: bold;
+  padding: 0.4rem 0.8rem;
+}
+.localgov_alert_banner--actions button:focus {
+  background: #410541;
+  color: #fff;
+ }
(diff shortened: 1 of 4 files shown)


css/localgov-alert-banner.css
```diff
@@ -9,14 +9,30 @@
...
   display: flex;
   justify-content: space-between;
   align-items: flex-end;
 }
+.localgov_alert_banner--actions button {
+  background: #77bb55;
+  border: none;
+  color: #000;
+  font-weight: bold;
+  padding: 0.4rem 0.8rem;
```
andybroomfield, line 25: Is adding the button colours intentional here? That would be better placed for a theme to handle. If our aim is to have the button not look like a system button then it might be a good idea to add the following here and do a basic style like a link. If not best to leave as a button and in the theme apply the right styling. 
````css
-webkit-appearance: none;
-moz-appearance: none;
appearance: none;
````
I

danchamp: Thanks @andybroomfield that makes sense. Have removed the button styling and left with system styling, looking at the localgov_theme I can now see that's where this should be handled.

Issue #65, opened 2020-07-08
Update paths for Add / Edit operations


ekes:
> I notices the path includes .../entity-type/bundle/id (bundle not being necessary and actually being a bit confusing on breadcrumb to path); 
> but more important shouldn't the path of the view list be the same as the one created for the collection-list to override it? Or is there a plan to put it somewhere/change the collection-list path?

andybroomfield:
Should follow the same pattern as nodes and live under the alert banners admin view path.
/admin/content/alert-banners/[bundle]/[id]/[operation]
eg.
/admin/content/alert-banners/add
/admin/content/alert-banners/add/localgov-alert-banner/
/admin/content/alert-banners/localgov-alert-banner/1
/admin/content/alert-banners/localgov-alert-banner/1/edit 
etc...

ekes:
The node pattern is:-

/node/[bundle]/add
but once it exists it's
/node/[id]/[operation]
The admin collection-list being
/admin/content
The non-admin collection is
/node

As this is 'content' and it's not content with a non-admin stand-alone view mode:

admin collection-list:
/admin/content/alert-banners
individual view:
/admin/content/alert-banners/[id]
operations:
/admin/content/alert-banners/[id]/[operation]
add types list:
/admin/content/alert-banners/add
add:
/admin/content/alert-banners/add/[bundle]

This is the same pattern as the stand alone (also admin view) 'paragraphs' does.

It also avoids using the partial path part /admin/content/alert-banners/[bundle] which doesn't actually have anything on it. Unless you plan to have separate lists of content for each bundle.

(and the node-type pattern is collection /admin/structure/types ; /admin/structure/types/add; /admin/structure/types/[bundle]/[operation])

andybroomfield:
@ekes Ok, that makes more sense I guess. I don't think many sites will have different bundles so they could just have a filter in the admin view to sort types (How content dashboards work will be another issue I'll raise on main localgov projects).

I'm happy with the above pattern, basically copy what nodes does but sit under the admin dashboard path.

PR #68, opened 2020-07-09
Feature/56 confirmation

ekes:
Adds the confirmation page.
Adds link to the admin list view.

Also moves the block to optional config. As it can be installed without the theme (allows tests to run without the theme, and it is possible to install separately and place the block).
Also changes the admin list path to /admin/contents/alert-banner in line with the rest of the other paths #65 (and actually now too)


ekes:
> The confirmation does not appear if you tick the publish box on the edit form of the entity. This is the same behaviour as before when using flag, would it be possible in the future to have it redirect to the confimration page when ticking that box?

I was expecting more that you'd remove the published checkbox from the form; and if there is no destination on form submission (the likely other destination is the view list) redirect to the status (make live) page.

> Its currently doing a page redirect, assume the modal version will come in a later PR?

If it's still desired to make it modal rather than a url that can be added to the link in the view. Obviously it's still actually now a form and a tab on the entity itself as well.

> The paths use underscores in the name and not hypens, is that the drupal way?

The only path this adds is `{entity.canonical}/status` the status-form for changing the published state. I think all the paths with underscores (this the additional `localgov_alert_banner` bit of the paths?) are part of my motivation for making #65

andybroomfield:
@ekes Thanks, more my miusunderstanding as to what is in this PR.
I'm happy for this to be merged and the paths be picked up as a new PR in that case.
And a new issue for the modal / if we should remove publish from the form (I'd prefer to keep it there as emergency publishers may be needing to write these banners in an emergency, but can be discussed the best implementation).

ekes:
> And a new issue for the modal / if we should remove publish from the form (I'd prefer to keep it there as emergency publishers may be needing to write these banners in an emergency, but can be discussed the best implementation).

The flow for creation of banners would become.

Go direct to the create form. 
Submit. 
Get redirected to {banner-path}/status where you get a preview and asked if you want to 'set this live?'

Otherwise you are on the list of banners. Where there are the 'set this live' 'remove this' links on existing banners. And the linkt to create new, flow above.

PR #69, opened 2020-07-10
Basic styling for default alert types

danchamp:
This PR provides basic styling for each of the default alert types, based on [GDS examples](https://docs.publishing.service.gov.uk/manual/emergency-publishing.html#death-of-a-notable-person). It also makes a minor adjustment to the template to move type classes to the parent div.

Styles display as:

announcement
minor
major
death

### DIFF
--- a/css/localgov-alert-banner.css
+++ b/css/localgov-alert-banner.css
@@ -1,7 +1,44 @@
+/**
+ * @file
+ * Style rules for alert banners
+ */
+
+/* Default */
 .localgov_alert_banner {
-  background-color: #fc3;
-  color: #000;
+  background-color: #00856A;
 }
+.localgov_alert_banner,
+.localgov_alert_banner a {
+  color: #fefefe;
+}
+.localgov_alert_banner button {
+  color: #121212;
+}
+
+/* Announcement */
+.localgov_alert_banner--announcement {
+  background-color: #00856A;
+}
+
+/* Minor */
+.localgov_alert_banner--minor {
+  background-color: #fc3;
+  color: #121212;
+}
+.localgov_alert_banner--minor a {
+  color: #121212;
+}
+
+/* Major */
+.localgov_alert_banner--major {
+  background-color: #C00010;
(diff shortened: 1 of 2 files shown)


css/localgov-alert-banner.css
```diff
@@ -1,7 +1,44 @@
...
 .localgov_alert_banner {
+  background-color: #00856A;
+}
+.localgov_alert_banner,
+.localgov_alert_banner a {
+  color: #fefefe;
+}
+.localgov_alert_banner button {
+  color: #121212;
+}
```
andybroomfield, line 16: Is this to stop the button appearing as white with the localgov theme?
That might be better as an adjustment there rather than in module?
(Just a thought, this is fine for now).

danchamp: It's to override the default color value (#fefefe) which would otherwise cascade to the button.

This could be solved by setting a default with a light background and dark color, so the explicit button color wouldn't be needed, or not setting a default at all, since the base alert banner must have a type?

andybroomfield: @danchamp Do you think
````css
.localgov_alert_banner button {
  color: initial;
}
````
Is widely supported? This sets it back to the system style sheet. Just feels a bit out of place setting a colour for the button.

andybroomfield: Not IE11 which may be a deal breaker (still widly deployed in councils, though Chrome is default)
https://caniuse.com/#feat=css-initial-value

danchamp: I agree what's there is somewhat clumsy, `initial` would be a neat solution.

Not sure about what level of support we should be giving to IE11, maybe something to be discussed with @finnlewis if it's not already defined?

danchamp: Just found the GDS browser support list:

https://www.gov.uk/service-manual/technology/designing-for-different-browsers-and-devices

Which does include IE11.

Issue #70, opened 2020-07-10
Banner create and set live flow


ekes:
> And a new issue for the modal / if we should remove publish from the form (I'd prefer to keep it there as emergency publishers may be needing to write these banners in an emergency, but can be discussed the best implementation).

The flow for creation of banners would become.

Go direct to the create form. 
Submit. 
Get redirected to {banner-path}/status where you get a preview and asked if you want to 'set this live?'

Otherwise you are on the list of banners. Where there are the 'set this live' 'remove this' links on existing banners, if the form is displayed in a modal or a link as now is an enhancement. And there is the link to create new, flow above.

_Originally posted by @ekes in https://github.com/localgovdrupal/localgov_alert_banner/pull/68#issuecomment-656637742_

andybroomfield:
I will check this with BHCC comms team to see if it works ok for them.
At present in BHCC, the put live uses the checkbox on the banner itself (although we also shortcut the admin screens to only ever editing a single banner).
The main concern is to avoid user confusion when a banner needs to be put live quickly, whilst still having some confirmation to prevent a banner going live if its just being drafted for future use.
Also would the extra step apply even if a user was not intending to put the banner live, but just writing one for future use? My current thoughts are that I think it should only trigger the confirmation step when there is explicit user confimration that they wish to put the banner live, otherwise it should go back to the list of alert banners.

ekes:
> Also would the extra step apply even if a user was not intending to put the banner live, but just writing one for future use? My current thoughts are that I think it should only trigger the confirmation step when there is explicit user confimration that they wish to put the banner live, otherwise it should go back to the list of alert banners.

Actually that would be simple enough:-

Remove the published checkbox from the form.
Place a non-field checkbox on the form that says 'I want to set this live directly'.
On submission of the entity form if set the destination based on the value of that checkbox.

ekes:
So I wasn't intending on doing this directly, but before I closed the files quickly: https://github.com/localgovdrupal/localgov_alert_banner/commit/daaca478d246c262c439218e0d58b44664e72abb

No tests. So if someone wants to write them that would be great.
Also it redirects to the .collection list which is currently the coded generated one for the bundle. That would be updated and made nicer by getting rid of the individual per-bundle ones, and overriding the entity one, which is the paths #65 issue.

andybroomfield:
Discovered this doesn't override the destination parementer so I will add fix for that.

andybroomfield:
[...]

PR #71, opened 2020-07-15
Redirect to confirmation step when changing state on the form

andybroomfield:
Ref #70

When user saves an alert banner and the state is changed, users get redirected
to the confirmation step.
This occurs even if there is a destination paremeter.

### DIFF
--- a/localgov_alert_banner.info.yml
+++ b/localgov_alert_banner.info.yml
@@ -8,6 +8,7 @@ dependencies:
   - drupal:field
   - drupal:field_group
   - drupal:link
+  - drupal:node
   - drupal:options
   - drupal:user
   - drupal:views
--- a/tests/src/Functional/AlertConfirmationTest.php
+++ b/tests/src/Functional/AlertConfirmationTest.php
@@ -107,8 +107,8 @@ class AlertConfirmationTest extends BrowserTestBase {
     $this->assertSession()->addressEquals($alert->toUrl('status-form')->toString());
     $this->getSession()->getPage()->pressButton('Confirm');
 
-    // Do not change the banner state and verify that is does not go to the
-    // confirmation page.
+    // Do not change the banner state and verify that user is not redirected to
+    // the confirmation form page.
     $edit_url = $alert->toUrl('edit-form')->toString();
     $form_vars = [
       'status-change' => 0,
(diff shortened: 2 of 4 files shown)


andybroomfield:
This also adds tests for the redirection.
TODO: Fix the redirect to go to the alert banner admin view instead of the collection (that could be a new PR).

tests/src/Functional/AlertConfirmationTest.php
```diff
@@ -63,6 +63,7 @@ public function testAlertConfirmation() {
     $this->getSession()->getPage()->pressButton('Confirm');
     $this->assertSession()->pageTextContains('The alert banner ' . $title . ' has been set live');
 
+    $this->drupalGet($alert->toUrl('canonical')->toString());
```
ekes, line 66: I'm fine with this. I kinda deliberately didn't put the url in to allow it to go to any alert/{id}/{action} as long as it has the feedback and the tab to remove.

localgov_alert_banner.info.yml
```diff
@@ -8,6 +8,7 @@ dependencies:
   - drupal:field
   - drupal:field_group
   - drupal:link
+  - drupal:node
```
ekes, line 11: Not important, I think it's pretty rare not to have node, but does this depend on it?

tests/src/Functional/AlertConfirmationTest.php
andybroomfield: Removing it gives this error
>  ✘ Alert confirmation
   │
   │ The link Remove banner was not found on the page.
   │ Failed asserting that an array has the key 0.
   │ 
   │ /app/web/core/tests/Drupal/Tests/UiHelperTrait.php:496
   │ /app/web/modules/contrib/localgov_alert_banner/tests/src/Functional/AlertConfirmationTest.php:67
   │ 

Does that mean the test is broken or that the page they land on should have the remove link?
I think at the monent saving get redirected back to the collection form.

localgov_alert_banner.info.yml
andybroomfield: Required by this in AlertBannerForm.php
````php
// Create the vertical tabs like on node edit forms.
// @src https://drupal.stackexchange.com/a/276907
$form['#theme'][] = 'node_edit_form';
$form['#attached']['library'] = ['node/drupal.node'];
````

andybroomfield:
> TODO: Fix the redirect to go to the alert banner admin view instead of the collection (that could be a new PR).
This will be a new PR.

andybroomfield:
Is this ok to merge @ekes @finnlewis ?

finnlewis:
I've not tested it, but the code looks good to me and the tests pass, so no objections from me.

PR #79, opened 2020-08-10
Replace the alert banner display view with a custom block

### LINKED ISSUE #42: Replace views block with custom block for displaying the alert banner
Currently views is used to display the alert banner to end users.
Since only one banner at a time is displayed, it may be less resource intensive to simply use an entityquery in a custom block.
This would remove a lot of extra markup that gets added, though it has the cost of making the feature a little less customisable (users who wished to expanded the feature would have to rebuild a view).

andybroomfield:
Fix #42

Basic custom block plugin that displays the published alert banner.
This replaces the view previously used, so the markup is more streamlined.

Should not require a new test as this will use the existing check that 
alert banner is visible test (`AlertBannerBlockTest`).


Adnan-cds:
Code changes look okay and clearly provides me with an alert banner.  But then I am left with two similar alert banners one of which came from the master branch.  This is not a problem for a test site like mine, but when these changes reach the [UAT site](https://localgovdrupal.agile.coop/), it will also have the same issue.  An update hook similar to the following can save us from such troubles:
```
function localgov_alert_banner_update_N() {
  Drupal::service('entity_type.manager')->getStorage('view')->load('alert_banner')->delete();
}
```

andybroomfield:
@Adnan-cds See localgovdrupal/localgov#79 Re: Update hooks.

andybroomfield:
Assuming with the issue around update hooks resolved, we can proceed with the merge?

Adnan-cds (approved):
I am happy with this change.

As discussed in https://github.com/localgovdrupal/localgov/issues/79, we don't need update hooks right now.

Issue #86, opened 2020-10-10
Coding standards - failing on 'Call to deprecated constant REQUEST_TIME'


finnlewis:
See: https://github.com/localgovdrupal/localgov/runs/1232632181#step:6:27
```
 ------ ---------------------------------------------------------------------------------------- 
  Line   modules/contrib/localgov_alert_banner/src/Form/AlertBannerEntityRevisionRevertForm.php  
 ------ ---------------------------------------------------------------------------------------- 
  132    Call to deprecated constant REQUEST_TIME: Deprecated in drupal:8.3.0                    
         and is removed from drupal:10.0.0. Use                                                  
         \Drupal::time()->getRequestTime();                                                      
 ------ ---------------------------------------------------------------------------------------- 

 ------ --------------------------------------------------------------------------------------------------- 
  Line   modules/contrib/localgov_alert_banner/src/Form/AlertBannerEntityRevisionRevertTranslationForm.php  
 ------ --------------------------------------------------------------------------------------------------- 
  92     Call to deprecated constant REQUEST_TIME: Deprecated in drupal:8.3.0                               
         and is removed from drupal:10.0.0. Use                                                             
         \Drupal::time()->getRequestTime();                                                                 
 ------ --------------------------------------------------------------------------------------------------- 
```

PR #87, opened 2020-10-10
Fix coding standards discussed in #86.

finnlewis:


finnlewis:
Looks like the coding standards are now passing. When the tests finish, @Adnan-cds perhaps you could approve?

src/Form/AlertBannerEntityRevisionRevertForm.php
```diff
@@ -129,7 +129,7 @@ public function submitForm(array &$form, FormStateInterface $form_state) {
   protected function prepareRevertedRevision(AlertBannerEntityInterface $revision, FormStateInterface $form_state) {
     $revision->setNewRevision();
     $revision->isDefaultRevision(TRUE);
-    $revision->setRevisionCreationTime(REQUEST_TIME);
+    $revision->setRevisionCreationTime(\Drupal::time()->getRequestTime());
```
andybroomfield, line 132: Should we use dependancy injection here, or is this ok?

andybroomfield (approved):
Happy for these to merge.

src/Form/AlertBannerEntityRevisionRevertForm.php
finnlewis: Sorry, I merged on seeing your approval and then saw your comment. How does dependancy injection work? Can you provide an example?

Adnan-cds: Hi Finn,
Within the create() [method](https://github.com/localgovdrupal/localgov_alert_banner/blob/00d83bfb6c7368771e7387a6d7abe0502f249786/src/Form/AlertBannerEntityRevisionRevertForm.php#L42), you can add the 'datetime.time' service as a class method (say $instance->dateTime).  The REQUEST_TIME constant can then be replaced with ```$this->dateTime->getRequestTime()```.  But don't worry, I will raise another pull request give that this has been merged already.

andybroomfield: In the past I've got deprecation warnings for 'dont use \Drupal methods in classes' however the tests pass.

Adnan-cds: I have added a [pull request](https://github.com/localgovdrupal/localgov_alert_banner/pull/88) to use dependency injection when inserting the request time.  I will assign that to both of you for review if it passes the tests :)

@andybroomfield @finnlewis

PR #90, opened 2020-10-16
Fix for typo in the label for announcement type

Adnan-cds:
A not so significant change :)

Spotted it while setting up the Croydon site.

### DIFF
--- a/config/install/field.storage.localgov_alert_banner.type_of_alert.yml
+++ b/config/install/field.storage.localgov_alert_banner.type_of_alert.yml
@@ -14,7 +14,7 @@ settings:
   allowed_values:
     -
       value: 00--announcement
-      label: Annoucement
+      label: Announcement
     -
       value: 20--minor
       label: 'Minor alert'


Adnan-cds:
Hi Andy,
I realized this is a config change will need an accompanying update hook implementation.  Let's leave it to rest now.  I will try to added an update hook in a week or so.

finnlewis (approved):
I think this is fine @Adnan-cds 

This might be a config change, but it would not break anything if it was or wasn't implemented on a site - there is not likely to be any dependency chain on this label, right?

Adnan-cds:
> there is not likely to be any dependency chain on this label, right?

Nah, there shouldn't be any dependency.  It's just that the existing sites will continue to display the typo.

andybroomfield:
This is fine, since we have to assume that a site might change these labels (we can communicate the change to the group for those that havn't caught the misspelling).

Adnan-cds:
Great.  I am merging it then.

I won't raise any follow-up pull request for a new **release**.  A fixed typo in a label probably doesn't merit a new release.

PR #92, opened 2020-11-19
Fix/91 title callback parameters

ekes:
@finnlewis To check that this works when installed with easybreadcrumb that was triggering the error when getting the title.
@andybroomfield To sanity check it.

### DIFF
--- a/src/AlertBannerEntityHtmlRouteProvider.php
+++ b/src/AlertBannerEntityHtmlRouteProvider.php
@@ -47,7 +47,7 @@ class AlertBannerEntityHtmlRouteProvider extends AdminHtmlRouteProvider {
     }
 
     if ($settings_form_route = $this->getSettingsFormRoute($entity_type)) {
-      $collection->add("$entity_type_id.settings", $settings_form_route);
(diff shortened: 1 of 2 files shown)


tanc:
@ekes I've tested this by patching locally and can confirm it fixes the problem: I can now make alerts 'live' with no errors.
Thanks!

andybroomfield (approved):
I have tested this in both scenarios with and without easy breadcrumb installed and this works.

I'd be interested to know what was causing the failure with Easy Breadcrumb and if its possible for the tests to be altered to catch this.

ekes:
In EasyBreadcrumbBuilder::build() it did `$title = $this->titleResolver->getTitle($route_request, $route_match->getRouteObject());` And it was calling the `\Drupal\Core\Controller\TitleResolverInterface` this way that called the _title_callback via the Symfony `RequestAttributeValueResolver` which assumes the parameter of the function is the key value in the request (this has been set by Drupal `RouteMatchValueResolver` on the whole process being triggered previously.

So I guess the way to test it. Would be to capture what's in the $route_request and $route_match when correctly run, and run the title resolver?

Issue #89, opened 2020-10-15
Decide what to do with release tag mix up.


andybroomfield:
Sort most recent release is currently v1.0.1-alpha1. But the Drupal pattern should be v.1.0.0-alpha2.
We can't release v.1.0.0-alpha2 as that will not upgrade those with 1.0.1.
If we delete this release, others who have updgraded will need to reinstall the package.
This might be ok as not many installs now, but we are also close to a v1.0 release anyway we could get this through beta and to v1.0.1 quickly for it not to matter?

Thoughts.

finnlewis:
@ekes and I were discussing this today, we think it will be fine.

We've removed the 1.0.1-alpha1 tag from Packagist and Github, which should result in no changes as it was the same as 1.0.0-alpha1.

So now we should be good to continue the increments

1.0.0-alpha1
1.0.0-alpha2
1.0.0-alpha3

up to (if needed)

1.0.0-beta1
1.0.0-beta2
...

Then 

1.0.0

andybroomfield:
@finnlewis @ekes Thanks, if your happy with that then please go ahead.

PR #94, opened 2020-11-20
Merge to release branch and create new release tag for 1.0.0-alpha3

finnlewis:
This is a pull request to review recent commits before tagging 1.0.0-alpha3

Each commit will already have been reviewed in a pull request, so this should be just a sanity check.

Comments welcome, or just a "looks good to me"!

### DIFF
--- a/config/install/field.storage.localgov_alert_banner.type_of_alert.yml
+++ b/config/install/field.storage.localgov_alert_banner.type_of_alert.yml
@@ -14,7 +14,7 @@ settings:
   allowed_values:
     -
       value: 00--announcement
-      label: Annoucement
+      label: Announcement
     -
       value: 20--minor
       label: 'Minor alert'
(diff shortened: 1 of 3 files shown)


Adnan-cds (approved):
looks good to me"! :D

andybroomfield:
Did we resolve #89?

finnlewis:
Just cross posting with #89 

@ekes and I were discussing this today, we think it will be fine.

We've removed the 1.0.1-alpha1 tag from Packagist and Github, which should result in no changes as it was the same as 1.0.0-alpha1.

So now we should be good to continue the increments

1.0.0-alpha1
1.0.0-alpha2
1.0.0-alpha3

up to (if needed)

1.0.0-beta1
1.0.0-beta2
...

Then 

1.0.0

ekes:
There is already an 1.0.0-alpha2 but it was not getting pulled as 1.0.1-alpha1 was higher. 
Deleting 1.0.1-alpha1 at packagist (manually) has made 1.0.0-alpha2 the current version.

andybroomfield (approved):
Happy for this release to go ahead with new tags.

PR #101, opened 2020-12-14
Remove tests dependancy on localgov_profile

### LINKED ISSUE #80: Rewrite test to not require localgov install profile
Currently the test for checking if an alert banner displays (`AlertBannerBlockTest`) relies on installing the localgov profile.

Instead, the test should place and confiqure the block created in #79.

andybroomfield:
Fix #80

Remove tests dependancy on localgov_profile so they can be run with just this
module.
Use place block rather than installing the profile and localgov theme


andybroomfield:
@finnlewis If the tests pass, as its minor would it be ok to just merge?

finnlewis:
Sure thing... looks trivial enough. 👍

PR #104, opened 2021-01-05
Reset tests back to php 7.4

andybroomfield:


andybroomfield:
Going to merge as this just resets the tests back to 7.4 as they are now working.

PR #109, opened 2021-01-25
Make banner block type display configurable.

j4-m:
* Adds a block form which allows admins to choose the alter banner types they want the block to display
* Use the configuration to filter the queried alter banners before render.

### DIFF
--- a/src/Plugin/Block/AlertBannerBlock.php
+++ b/src/Plugin/Block/AlertBannerBlock.php
@@ -161,7 +161,7 @@ class AlertBannerBlock extends BlockBase implements ContainerFactoryPluginInterf
   protected function mapTypesConfigToQuery() : array {
     $include_types = $this->configuration['include_types'];
     return array_filter($include_types, function ($t) {
-      return $t !== 0;
+      return (bool) $t;
     });
   }
 
--- a/src/Plugin/Block/AlertBannerBlock.php
+++ b/src/Plugin/Block/AlertBannerBlock.php
@@ -2,6 +2,7 @@
 
 namespace Drupal\localgov_alert_banner\Plugin\Block;
 
+use Drupal\Core\Form\FormStateInterface;
 use Drupal\Core\Plugin\ContainerFactoryPluginInterface;
 use Symfony\Component\DependencyInjection\ContainerInterface;
 use Drupal\Core\Entity\EntityTypeManagerInterface;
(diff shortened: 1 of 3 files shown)


andybroomfield:
Thanks @iam-jam 
Do you know why the tests are failing, saying missing schemas?

iam-jam:
Moving to WIP as we'll need to update the tests to cover the new configuraion.

andybroomfield:
Thanks for your work so far @iam-jam. 
I'm seeing an error on the block configuration screen on first install.
> Error message
Warning: Invalid argument supplied for foreach() in Drupal\Core\Render\Element\Checkboxes::valueCallback() (line 100 of core/lib/Drupal/Core/Render/Element/Checkboxes.php).

iam-jam:
No worries @andybroomfield. As of the latest commits, there shouldn't be any errors when saving the configuration form and hopefully the schema issue has been resolve too... just need add some test coverage now.

andybroomfield (changes requested):
Thanks for your efforts on this @iam-jam.

I'm still seeing the following error on the block configuration page when first installing the module.
> Error message
Warning: Invalid argument supplied for foreach() in Drupal\Core\Render\Element\Checkboxes::valueCallback() (line 100 of core/lib/Drupal/Core/Render/Element/Checkboxes.php).

To replicate, do a fresh install and enable localgov_alert_banner. Then go to the block configuration screen without having placed a block.
The error will disapear once I save the block for the first time.

Screenshot 2021-01-29 at 1 45 05 pm

Additonal amends:
- Can we change the label Display types to Banner types to match the admin view.
- Would it be possible to have the negate this condition option added

finnlewis:
I can confirm the same error. 
I can't immediately see why the default_value key is just "" when it is expecting an array. 
If you can show me how to fix this, I can use it as a debugging demo at OxDUG!

src/Plugin/Block/AlertBannerBlock.php
```diff
@@ -65,6 +66,46 @@ public static function create(ContainerInterface $container, array $configuratio
...
+    $config_options = [];
+    foreach ($type_storage->loadMultiple() as $id => $type) {
+      $config_options[$id] = $type->label();
+    }
+    $form['include_types'] = [
+      '#type' => 'checkboxes',
+      '#options' => $config_options,
+      '#title' => $this->t('Display types'),
+      '#description' => $this->t('If no types are selected all will be displayed.'),
+      '#default_value' => !empty($config['include_types']) ? $config['include_types'] : '',
```
andybroomfield, line 94: @finnlewis @iam-jam I think this might be what is causing the error.
I think default value should be an array (even if it is null).

iam-jam: Yep that's it @andybroomfield. I had actually spotted this before but forgot to fix it. Shame this hasn't been picked up in tests.

finnlewis:
Yes - I got the same place following the debugger and the code cahnges :)

andybroomfield:
Hi @j4-m Thanks for your efforts so far. I'm stlll getting the error, did we get anywhere with this?

src/Plugin/Block/AlertBannerBlock.php
andybroomfield: Just to confirm that changing 
`'#default_value' => !empty($config['include_types']) ? $config['include_types'] : '',`
to
`'#default_value' => !empty($config['include_types']) ? $config['include_types'] : [],`
resolves this issue. If we can get this change added we can get this in a new release.

finnlewis: Ditto. @j4-m could you commit the change to your forked branch?

j4-m:
@andybroomfield @finnlewis the amendment to the default value has just gone in.

andybroomfield (approved):
Error now fixed so clear to merge @j4-m

PR #111, opened 2021-02-02
Feature/110 Add content local task for alert banners

tebb:
I have added the 'Alert banners' local task and tested that it appears after routing and links cache is flushed.

I added a second commit which (I believe) corrects some capitalisation and a typo.  I wasn't sure if it's OK to do that in this specific issue.

### DIFF
--- a/localgov_alert_banner.links.task.yml
+++ b/localgov_alert_banner.links.task.yml
@@ -31,4 +31,4 @@ entity.localgov_alert_banner.delete_form:
 entity.localgov_alert_banner.collection:
   route_name: view.localgov_admin_manage_alert_banners.localgov_alert_banner_admin_list
   base_route:  system.admin_content
-  title: 'Alert Banners'
+  title: 'Alert banners'
--- a/localgov_alert_banner.links.task.yml
+++ b/localgov_alert_banner.links.task.yml
@@ -16,14 +16,19 @@ entity.localgov_alert_banner.version_history:
   title: 'Revisions'
 
 entity.localgov_alert_banner.status_form:
-  route_name:  entity.localgov_alert_banner.status_form
-  base_route:  entity.localgov_alert_banner.canonical
+  route_name: entity.localgov_alert_banner.status_form
+  base_route: entity.localgov_alert_banner.canonical
   title: 'Status'
   weight: 10
   class: '\Drupal\localgov_alert_banner\Plugin\Menu\LocalTask\StatusFormTab'
 
 entity.localgov_alert_banner.delete_form:
-  route_name:  entity.localgov_alert_banner.delete_form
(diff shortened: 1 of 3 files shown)


andybroomfield:
@tebb is this reviewable?

tebb:
Thanks @andybroomfield - Yes please.

(We thought that creating pull requests automatically queued things for review.  Should we be requesting code reviews?)

andybroomfield (changes requested):
I've installed this and it worked fine at first, however the tabs are missing when a user goes to the manage alert banners page.
Will post screenshots shortly.

localgov_alert_banner.links.task.yml
```diff
@@ -16,14 +16,19 @@ entity.localgov_alert_banner.version_history:
...
 entity.localgov_alert_banner.delete_form:
-  route_name:  entity.localgov_alert_banner.delete_form
-  base_route:  entity.localgov_alert_banner.canonical
+  route_name: entity.localgov_alert_banner.delete_form
+  base_route: entity.localgov_alert_banner.canonical
   title: Delete
   weight: 10
+
+entity.localgov_alert_banner.collection:
+  route_name: view.localgov_admin_manage_alert_banners.localgov_alert_banner_admin_list
```
andybroomfield, line 32: Can we change the route name to `entity.localgov_alert_banner.collection`
Theres a weird bug that when you go to the alert banner page the tabs no longer show. I think its to do with defining a collection route in the entity class, but we don't use the collection, we use the view. This causes issues with the menu / tab display but we will sort these out in another issue.

localgov_alert_banner.links.task.yml
```diff
@@ -16,14 +16,19 @@ entity.localgov_alert_banner.version_history:
...
+  base_route:  system.admin_content
```
andybroomfield, line 33: Remove the extra space
`base_route: system.admin_content`

localgov_alert_banner.links.task.yml
```diff
@@ -16,14 +16,19 @@ entity.localgov_alert_banner.version_history:
...
+  title: 'Alert banners'
```
andybroomfield, line 34: We need to add `weight: 50` to the bottom so it shows last in list.

andybroomfield (changes requested):
Thanks for your work @tebb and apologies for late reviewing.
I had a look and overall good, however there are some quirks I spotted with the tabs not showing once I go to the manage alert banner page. See inline comments.

andybroomfield:
I've added the fixes, if theres anyone who can give this a review before I merge.
Then I can get a release prepped.

danchamp (approved):
Looks good, tested and working as expected.

andybroomfield:
Thanks @danchamp

PR #123, opened 2021-04-28
format templates to follow Drupal practices

### LINKED ISSUE #122: Update Frontend to follow Drupal practices
Update 
- template to follow drupal practices
- CSS to follow BEM
- JS to use new classes
- localgov_alert_banner.module to use new template suggestion

You can see a video of my thought process here:
https://www.twitch.tv/videos/1003684550

That will be deleted after 14 days automatically by Twitch, so I'll upload a version of it to YouTube later when it's finished processing on my computer.

Update: YouTube is here:
https://www.youtube.com/watch?v=PtO3ur2PUUY

markconroy:
Closes #122


templates/localgov-alert-banner.html.twig
```diff
@@ -0,0 +1,68 @@
...
+{% set classes = [
+    'js-localgov-alert-banner',
+    'localgov-alert-banner',
+    type_of_alert ? 'localgov-alert-banner--' ~ type_of_alert,
+    is_front ? 'localgov-alert-banner--homepage',
+    has_link ? 'localgov-alert-banner--has-link' : 'localgov-alert-banner--no-link'
+  ]
+%}
+
+<article {{ attributes.addClass(classes) }} role="alert">
```
andybroomfield, line 35: Are we sure we want to use `<article>` here?
An alert banner isn't really an article, though it might point to one, or it could just be a general alert saying forms are down.

css/localgov-alert-banner.css
```diff
@@ -4,52 +4,74 @@
  */
 
 /* Default */
-.localgov_alert_banner {
+.localgov-alert-banner {
+  padding-top: 1rem;
+  padding-bottom: 1rem;
```
andybroomfield, line 9: Not sure about the extra padding. Given alert banners tend to be compacted when viewed (see examples on Croydon.gov.uk etc...) then this would be a style that almost every council has to override. May be better to remove it as default.

templates/localgov-alert-banner.html.twig
markconroy: An `<article>` in HTML is also not just an "article" like you would expect in a newspaper (it is that, and more). It's a "self-contained composition". Article tag is also used for our media items (and other things) in Drupal.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/article

These alerts are self-contained. They can display on any page and do not lose their meaning depending on the page they are on. They also don't lose their meaning if they were standalone on a page with nothing else.

We can move it back to div if you wish.

css/localgov-alert-banner.css
markconroy: That padding is removed now.

templates/localgov-alert-banner.html.twig
andybroomfield: Ok, that makes sense. I take it we will do the same for all components of the site, so block will be getting the article markup too?

andybroomfield:
This is really good, thanks @markconroy. 
If we can get the tests passing again, I think it is just changing the class of several elements it looks for when checking if the banner is present and the hide button, then this should be good to merge.

templates/localgov-alert-banner.html.twig
markconroy: It would depend on the block to be honest. Not all blocks should be article, and it might not make sense for maintainability to create a template for every block type just so we can set article as the element.

But in cases where we _do_ have a template, then it makes sense to make sure we have the most appropriate element chosen.

markconroy:
@andybroomfield I've changed one test where I could see the class needed to change, but tests seem to be still breaking. Could you fix that by any chance?

andybroomfield:
Sometimes we have to re-run the tests if the tests have been changed, maybe something to do with paratest? @stephen-cox

andybroomfield (approved):

markconroy:
Looks like it passed.

@andybroomfield if you're okay with this now can I leave it to you to merge this for the release?

If anyone is using localgov_theme, they will need to update their CSS classes - I think the only one might be changing `localgov_alert_banner` to `localgov-alert-banner`.

Thanks a lot. I'm really happy with how this module has turned out. Great work.

Issue #124, opened 2021-04-28
Tag 1.0.0


ekes:
As per localgovdrupal/localgov#194 localgovdrupal/localgov#200

PR: template changes to be discussed not blocking
Issues: nothing blocking, mainly new features.

ekes:
@andybroomfield Do you want to tag this one? Is there anything I missed that wants doing before that?

andybroomfield:
@ekes I'll take it, I want to review some of the issues first but should be no problem to go ahead and tag.

ekes:
https://github.com/localgovdrupal/localgov_alert_banner/releases/tag/1.0.0

PR #127, opened 2021-04-29
Release branch ahead of tag 1.0.0

andybroomfield:
This is the first full release of Localgov_alert_banner
(Yes we're jumping past beta! See #124).

This represents the first stable version, so future releases will be using semantic versioning to manage breaking changes and new features.

Incorporates the template coding standards work thanks @markconroy.
This will require some updates to your theme css if you have customised the banner in your theme.
> If anyone is using localgov_theme, they will need to update their CSS classes - I think the only one might be changing localgov_alert_banner to localgov-alert-banner.

Also note that the template for the alert banner has changed from `localgov_alert_banner.html.twig`
to `localgov-alert-banner.html.twig ` 

@markconroy
format templates to follow Drupal practices
32b3ce8

@markconroy
update library definition to fix typo
0f20c34

@markconroy
remove padding from alert banner
95db581

@markconroy
update test to correct class name
21f6ba7


willguv:
Nice work @andybroomfield @markconroy!

finnlewis (approved):
Tests pass and it looks good to me!

Issue #120, opened 2021-04-21
LocalGov Alert Banner: specification for targetted alerts


finnlewis:
This issue is to help prepare and approve the specification for changes to the alert banner to support greater control over where alert banners are displayed on a site. 

## User story (example)

**As a** content designer

**I want to** publish an alert banner that only displays on /events section of the site

**So that** I can add a warning about COVID-19 safety at co-located events. 

## User story (example)

**As a** content designer

**I want to** publish an alert banner that only displays on news items

**So that** I can add a notice about our communications department.

finnlewis:
@andybroomfield @willguv 

Do we already have any defined specification than that on https://trello.com/c/ML8TaTYD/288-targeting-alerts-to-a-specific-page ?

Should we arrange a call to confirm where the main drivers for this are coming from?

willguv:
@finnlewis are we using the first regular story refinement session for this? Monday pm fortnightly? I think you were going to invite people. I can if not

andybroomfield:
> ## User story (example)
> **As a** content designer
> 
> **I want to** publish an alert banner that only displays on /events section of the site
> 
> **So that** I can add a warning about COVID-19 safety at co-located events.
> 

That can be achieved with existing functionlity.

1. Create a new alert banner type for event banners
2. Create a new alert banner block and select only display banners of type event, and remove event banners from the main block.
3. Set the conditions on the newly created events block to only display on the /events* path


> ## User story (example)
> **As a** content designer
> 
> **I want to** publish an alert banner that only displays on news items
> 
> **So that** I can add a notice about our communications department.

Same as above.

I think where it gets more complicated is if you have more complex conditions and don't want to have to manage new alert banner types and blocks for every scenario. I've been exploring using [condition field](https://www.drupal.org/project/condition_field) on the banner entity and had good results, although it did require adding a custom hook.

It does raise the question of how much functionality we want to actully support within the modules, and how much we would expect a council / orgnaisation to be able to build themselves. The reason we made this change to making alert banner an entity in its own right instead of just a text box was so that it can be leveraged with all of Drupals functionality. 
This doens't mean not to include it, just to raise it as something we genuinly expect the module to provide and to make sure we don't end up trying to build things that cover every use case.

finnlewis:
Just to note, update on the Trello card:  

> Ben Hills-Jones 31 minutes ago
[...]

PR #129, opened 2021-04-29
Experimental targeted banners using condition field (for preview purposes)

andybroomfield:
Add condition field support so that some banners can be limited to certain pages based on conditions plugins.

Ref #120 

Requires condition_field
  https://drupal.org/project/condition_field

To test, edit an alert banner type and add a condition field named `field_conditions`
Now you should be able to set up alert banner entities, and set visibility conditions on them.

This is a preview of potential condition support, so we need to decide if we will rely on a dependancy, or bake in conditions plugin support into the module.

### DIFF
--- a/localgov_alert_banner.module
+++ b/localgov_alert_banner.module
@@ -108,7 +108,7 @@ function localgov_alert_banner_theme_suggestions_localgov_alert_banner(array $va
 function localgov_alert_banner_localgov_alert_banner_view(array &$build, EntityInterface $entity, EntityViewDisplayInterface $display, $view_mode) {
 
   // Check if the field is present and has a value.
-  if ($entity->get('field_conditions')->isEmpty()) {
+  if (!$entity->hasField('field_conditions') || $entity->get('field_conditions')->isEmpty()) {
     return TRUE;
   }
 
--- a/localgov_alert_banner.module
+++ b/localgov_alert_banner.module
@@ -7,6 +7,9 @@
 
 use Drupal\Core\Entity\EntityInterface;
 use Drupal\Core\Routing\RouteMatchInterface;
+use Drupal\Core\Entity\Display\EntityViewDisplayInterface;
+use Drupal\condition_field\ConditionAccessResolver;
+use Drupal\user\Entity\User;
 
 /**
  * Implements hook_help().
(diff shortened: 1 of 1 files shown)


andybroomfield:
The tests for this do pass locally, not sure why they are failing here?

localgov_alert_banner.module
```diff
@@ -98,3 +101,61 @@ function localgov_alert_banner_theme_suggestions_localgov_alert_banner(array $va
...
+  // Bypass condition check if it is the alert banner route or confirmation.
+  $bypass_conditions = ($route_name == 'entity.localgov_alert_banner.canonical' || $route_name == 'entity.localgov_alert_banner.status_form');
+
+  if (!$bypass_conditions) {
+
+    // Loop through each condition.
+    foreach ($conditions_config as $condition_id => $values) {
+
+      // Add current node as context if required and on a node route.
+      $node = $route_parameters->get('node');
```
ekes, line 140: This got me to thinking. The conditions make sense, hopefully they'll fit the requirements that the product group come up with selecting where things are shown. But for this part...

Would it make more sense for this to be in the block build() https://github.com/localgovdrupal/localgov_alert_banner/blob/1a30057581758a56ced975c8cb11c7d9b952589a/src/Plugin/Block/AlertBannerBlock.php#L122
Good because it's earlier and also here you could get things like the node by adding it to the block context
```
 *   context = {
 *     "node" = @ContextDefinition("entity:node")
 *   }
```
and pulling it `$node = $this->getContextValue('node');`
Which gives the added bonus that iirc the ContextAwarePlugin stuff does the cache contexts?

andybroomfield: That could work, would there be circumstances we render the entity outside the block? Would we assume users who are doing that actully know what they are doing so should handle it themselves.

andybroomfield:
This work was done for preview purposes, though if there is a way of using the condition plugin without condition field it might help to remove the dependancy.

localgov_alert_banner.module
andybroomfield: Also, this block may still exist on non node pages. Does adding a block context of node break that @ekes ?

andybroomfield:
Closing this as condition field support is already merged.

PR #130, opened 2021-05-04
Split GitHub CI tests into separate jobs (localgovdrupal/localgov#179)

stephen-cox:
See: https://github.com/localgovdrupal/localgov/pull/178


stephen-cox:
Tests pass so merging.

PR #135, opened 2021-05-10
Delete AlertBannerRouteSubscriber.php

### LINKED ISSUE #93: Drupal\localgov_alert_banner\Routing\AlertBannerRouteSubscriber::alterRoutes()
I think this is legacy code that we've overlooked removing. It's not active now as it has an conditional on `entity.alert_banner.canonical` (why I noticed it checking for 'alert_banner' and not 'localgov_alert_banner' strings on #91 ).

andybroomfield:
Fix #93
Remove legacy code.


andybroomfield:
This should be a quick fix to remove some legacy close and close an old issue.

ekes (approved):

PR #132, opened 2021-05-05
Code fixes for Drupal 9

stephen-cox:
Closes #131

PR #139, opened 2021-05-14
1.0.1 release for Drupal 9 support

stephen-cox:
This is needed for Drupal 9 tests to start passing

Changelog:

#132 @stephen-cox Code Fixes for Drupal 9 support
#135 @andybroomfield Delete AlertBannerRouteSubscriber.php
#140 @stephen-cox  Updated Github CI to run 1.x branch against Drupal 8 and 9


andybroomfield:
@stephen-cox Are the tests fail a known issue and this will need merging to pass, or is this still under investigation?

stephen-cox:
This needs to be merged for other tests to pass, but the failure here is not expected; the install should complete, the other tests should fail on specific modules. I'll need to investigate

finnlewis (approved):
This looks good. Just the alert_banner CS checks as expected.

PR #140, opened 2021-05-17
Updated Github CI to run 1.x branch against Drupal 8 and 9

stephen-cox:
Part of https://github.com/localgovdrupal/localgov/issues/225

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,9 +2,11 @@ name: Test LocalGov Drupal Alert Banner
 
 on:
   push:
-    branches: [ master ]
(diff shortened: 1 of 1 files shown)


stephen-cox:
@andybroomfield The test failure here are expected as they refer to deprecated code in core. Otherwise everything is now passing.

If you're happy for this to be merged, we can then rerun the tests for the 1.0.1 PR and the tests should pass expect for then same deprecated code issues with localgov_core.

finnlewis (approved):
Looks good to me (and we agreed that we could review internally for this round of pull requests)

PR #144, opened 2021-06-04
Render all content fields on the alert banner

### LINKED ISSUE #143: Replace content.short_description in template with content|without('title', 'link')
I think it was this way before, but its since been changed (@markconroy ?)
This means the alert banner cannot have extra fields without a custom template.
Before I think we used `content|without('title', 'link')` so that any extra field gets printed.

I noticed this when experimenting with Layout Builder on the banner which wraps the fields in the content variable with the region and nothing was being printed out. The fix in the description fixes it.

andybroomfield:
Fix #143.
Use twig without filter to not print out the title or the link field, instead of just the description.

### DIFF
--- a/templates/localgov-alert-banner.html.twig
+++ b/templates/localgov-alert-banner.html.twig
@@ -46,7 +46,7 @@
         {% endif %}
         
         <div class="localgov-alert-banner__body">
-          {{ content|without('title', 'link' }}
+          {{ content|without('title', 'link') }}
         </div>
 
         {% if has_link %}
--- a/templates/localgov-alert-banner.html.twig
+++ b/templates/localgov-alert-banner.html.twig
@@ -45,11 +45,9 @@
           </h2>
         {% endif %}
         
-        {% if content.short_description.0['#text'] %}
-          <div class="localgov-alert-banner__body">
-            {{ content.short_description }}
-          </div>
-        {% endif %}
+        <div class="localgov-alert-banner__body">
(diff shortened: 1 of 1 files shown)


andybroomfield:
@markconroy @stephen-cox @Adnan-cds would be good to get this one checked and merged if possible.

Adnan-cds (approved):
Looks good to me :)

markconroy:
+1 from me

templates/localgov-alert-banner.html.twig
```diff
@@ -45,11 +45,9 @@
...
-          <div class="localgov-alert-banner__body">
-            {{ content.short_description }}
-          </div>
-        {% endif %}
+        <div class="localgov-alert-banner__body">
+          {{ content|without('title', 'link') }}
+        </div>
 
         {% if has_link %}
           <div class="localgov-alert-banner--content-link">{{ content.link }}</div>
```
markconroy, line 53: This class should be `.localgov-alert-banner__content-link` as an "element" of `.localgov-alert-banner` not a "modifier" of it.

andybroomfield: Shall I modify it here or as a new issue?
I assume there is some stylesheet in theme targeting this which is why it was left as is.
If I did change it here I would go with `localgov-alert-banner__link`

markconroy: Maybe leave it so, as it will not be available in the new theme that will launch soon.

PR #145, opened 2021-06-09
1.1.0 release

stephen-cox:
Adds conditional alert banner placement so banner display can be limited to specific paths.

Changelog:

#141 @stephen-cox Updated Github CI to run 1.x branch against Drupal 8 and 9
#142 @andybroomfield @stephen-cox Conditional alert banner placement
#144 @andybroomfield Render all content fields on the alert banner


stephen-cox:
@andybroomfield @ekes @finn Just checking we want a 1.1.0 release here and what the implications of this are.

The alert banner visibility work is a new feature so a 1.1.0 release seems appropriate. The only issue I see with this is the profile includes this as `~1.0.0`, locking this module as 1.0.x releases: https://github.com/localgovdrupal/localgov/blob/89291bb5902e4b4e213640bc8e1943d9a37d0917/composer.json#L22

Does this mean we need to do a new release of the profile as well? The visibility field is added to existing alert banners. Does this mean we need some sort of communications about this?

andybroomfield:
I'd say it is v1.1.0
I suppose whether it's just a patch realise for the profile or if we wait for a few more new features to up those release numbers.

andybroomfield:
Do we have a descision on this @finnlewis @ekes @Adnan-cds or shall we bring this up at technical group?

finnlewis:
v1.1.0 sounds good to me @andybroomfield @stephen-cox 

I think we go ahead and release this one, then talk about the profile update on a separate issue / pull request.

andybroomfield (approved):

andybroomfield:
This also includes #144 which alters the template to display all content fields.

PR #152, opened 2021-06-30
Move the cacheContexts and cacheTags to the alert banner entity

### LINKED ISSUE #150: Do we need the "session" cache context?
Hi Andy,
Our **cache_render** MySQL table is getting flooded with alert banner blocks.  It appears that each and every anonymous page request (that's not already in cache_page) is adding an alert banner block entry in cache_render.  This is because all anonymous page requests have different sessions and the alert banner is using the ["session" cache context](https://github.com/localgovdrupal/localgov_alert_banner/blob/ef0149e9b22b2dd92520a63876916a400a29b069/localgov_alert_banner.module#L38).  Do we really need the *session* cache context on Alert banner entities?

Please note that while there are only a few thousand pages in our site, the presence of the directory facets have created an unlimited combination of page URLs and page requests for those URLs.  That's what's causing the flooding :(

The **url.path:is_front** cache context is also contributing to the flooding due to a typo.  I have already raised a [pull request](https://github.com/localgovdrupal/localgov_alert_banner/pull/149) for that.

@andybroomfield

andybroomfield:
Fix #150

- Remove the session and user role contexts
- Move cache contexts to alert banner entity
- Move cache tags to alert banner entity

### DIFF
--- a/src/Entity/AlertBannerEntity.php
+++ b/src/Entity/AlertBannerEntity.php
@@ -392,9 +392,12 @@ class AlertBannerEntity extends EditorialContentEntityBase implements AlertBanne
    */
   public function getCacheContexts() {
 
+    $contexts = [];
+    $contexts = Cache::mergeContexts($contexts, ['url.path.is_front']);
+
     // Add cache contexts depending on the enabled visibility condition plugins.
     if ($this->hasField('visibility') && !$this->get('visibility')->isEmpty()) {
-      $contexts = [];
+
       $conditions_config = $this->get('visibility')->getValue()[0]['conditions'];
 
       foreach ($conditions_config as $condition_id => $values) {
@@ -406,7 +409,21 @@ class AlertBannerEntity extends EditorialContentEntityBase implements AlertBanne
       $this->addCacheContexts($contexts);
     }
 
-    return parent::getCacheContexts();
+    return Cache::mergeContexts(parent::getCacheContexts(), $contexts);
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getCacheTags() {
+
+    $cache_tags = [];
+
+    // Get token and use as a cache tag.
+    $token = $this->getToken();
+    $cache_tags = Cache::mergeTags($cache_tags, ['localgov.alert.banner.token:' . $token]);
+
+    return Cache::mergeTags(parent::getCacheTags(), $cache_tags);
   }
 
 }
--- a/src/Entity/AlertBannerEntity.php
+++ b/src/Entity/AlertBannerEntity.php
@@ -392,6 +392,7 @@ class AlertBannerEntity extends EditorialContentEntityBase implements AlertBanne
    */
   public function getCacheContexts() {
 
+    // Add front page context for front page classes.
     $contexts = ['url.path.is_front'];
 
     // Add cache contexts depending on the enabled visibility condition plugins.
@@ -406,11 +407,8 @@ class AlertBannerEntity extends EditorialContentEntityBase implements AlertBanne
       }
     }
 
-    // Add to the main cache contexts.
-    $this->addCacheContexts($contexts);
-
     // Return cache contexts.
-    return parent::getCacheContexts();
+    return Cache::mergeContexts(parent::getCacheContexts(), $contexts);
   }
 
   /**
@@ -420,7 +418,7 @@ class AlertBannerEntity extends EditorialContentEntityBase implements AlertBanne
 
     // Get token and use as a cache tag.
     $token = $this->getToken();
-    $cache_tags = ['localgov.alert.banner.token:' . $token];;
+    $cache_tags = ['localgov.alert.banner.token:' . $token];
 
     return Cache::mergeTags(parent::getCacheTags(), $cache_tags);
   }
--- a/src/Entity/AlertBannerEntity.php
+++ b/src/Entity/AlertBannerEntity.php
@@ -392,8 +392,7 @@ class AlertBannerEntity extends EditorialContentEntityBase implements AlertBanne
    */
   public function getCacheContexts() {
 
-    $contexts = [];
-    $contexts = Cache::mergeContexts($contexts, ['url.path.is_front']);
+    $contexts = ['url.path.is_front'];
 
     // Add cache contexts depending on the enabled visibility condition plugins.
     if ($this->hasField('visibility') && !$this->get('visibility')->isEmpty()) {
@@ -405,11 +404,13 @@ class AlertBannerEntity extends EditorialContentEntityBase implements AlertBanne
         $condition = $this->pluginManagerCondition->createInstance($condition_id, $values);
         $contexts = Cache::mergeContexts($contexts, $condition->getCacheContexts());
       }
-
-      $this->addCacheContexts($contexts);
     }
 
-    return Cache::mergeContexts(parent::getCacheContexts(), $contexts);
+    // Add to the main cache contexts.
+    $this->addCacheContexts($contexts);
+
+    // Return cache contexts.
+    return parent::getCacheContexts();
   }
 
   /**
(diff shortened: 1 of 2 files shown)


andybroomfield:
Initial fix.
Let me know if this works for you @Adnan-cds.

I was a little confused at the cacheContexts in the alertbanner entity. Should I be adding them via `$this->addCacheContexts($contexts);` or merging them with the parent tags as in `return Cache::mergeContexts(parent::getCacheContexts(), $contexts);`

Adnan-cds:
Code looks good Andy.  I will try it and get back to you.  Thanks a lot for the prompt action :)

src/Entity/AlertBannerEntity.php
```diff
@@ -392,9 +392,12 @@ public static function baseFieldDefinitions(EntityTypeInterface $entity_type) {
    */
   public function getCacheContexts() {
 
+    $contexts = [];
+    $contexts = Cache::mergeContexts($contexts, ['url.path.is_front']);
```
stephen-cox: I don't think this line is necessary. Cache contexts are just strings so you can initialise it with `$contexts = ['url.path.is_front'];`

src/Entity/AlertBannerEntity.php
```diff
@@ -406,7 +409,21 @@ public function getCacheContexts() {
       $this->addCacheContexts($contexts);
```
stephen-cox: As we're setting cache contexts outside the visibility field check I think this line should be moved out of the check block. Although this might not be needed at all.

src/Entity/AlertBannerEntity.php
```diff
@@ -406,7 +409,21 @@ public function getCacheContexts() {
...
     }
 
-    return parent::getCacheContexts();
+    return Cache::mergeContexts(parent::getCacheContexts(), $contexts);
```
stephen-cox, line 411: I'm not entirely sure of the proper way to do this, but what we're assuming here is that all the cache contexts for an alert banner are entirely defined in this method or on the parent. If we wanted to add cache tags to an alert banner somewhere else they could be added to the `cacheContexts` property using the `addCacheContexts` method, but this won't work any longer.

I think it's safer to add the new `$contexts` using to the entity using the `addCacheContexts` method and then either return the `cacheContexts` property or `parent::getCacheContexts()` (which will do the same).

It _may_ be better to actually move the logic here out of the `getCacheContexts` method and have it run by the constructor to ensure it's only run once. There's the possibility that it will be called multiple times, which won't do any harm, but is inefficient.

src/Entity/AlertBannerEntity.php
```diff
@@ -406,7 +409,21 @@ public function getCacheContexts() {
...
+  /**
+   * {@inheritdoc}
+   */
+  public function getCacheTags() {
+
+    $cache_tags = [];
+
+    // Get token and use as a cache tag.
+    $token = $this->getToken();
+    $cache_tags = Cache::mergeTags($cache_tags, ['localgov.alert.banner.token:' . $token]);
```
stephen-cox: This can be simplified to `$cache_tags = ['localgov.alert.banner.token:' . $token];`

src/Entity/AlertBannerEntity.php
```diff
@@ -406,7 +409,21 @@ public function getCacheContexts() {
...
+
+    return Cache::mergeTags(parent::getCacheTags(), $cache_tags);
```
stephen-cox, line 423: If this is called multiple times then a new `$token` will be generated each time. Might be better to move this to the constructor and add the cache tag using the `addCacheTags` method.

stephen-cox (changes requested):
@andybroomfield I've made a few comments. I'm not an expert on Drupal cache contexts and tags, so am happy to be overidden on some of the suggested changes. But having thought it through I'm not sure when I originally added code to the `getCacheContexts` method it was the best place for it. It may be better being called by the constructor. Happy to discuss further.

src/Entity/AlertBannerEntity.php
Adnan-cds: I thought the token doesn't change unless you update an alert banner.

src/Entity/AlertBannerEntity.php
Adnan-cds: > Although this might not be needed at all.

Seem to be correct.

src/Entity/AlertBannerEntity.php
andybroomfield: Yes the token won't change unless the banner is updated. 
Can we call methods in the constructor?

src/Entity/AlertBannerEntity.php
Adnan-cds: > If we wanted to add cache tags to an alert banner somewhere else they could be added to the cacheContexts property using the addCacheContexts method, but this won't work any longer.

As far as I understand, [parent::getCacheContexts()](https://git.drupalcode.org/project/drupal/-/blob/9.2.x/core/lib/Drupal/Core/Cache/CacheableDependencyTrait.php#L62) returns ```$cacheContexts```.  This means AlertBannerEntity::getCacheContexts() should always return whatever has been added through ```addCacheContexts()```

Adnan-cds:
Hi Andy,
I have finished my testing.  It was doing great until I added a Visibility condition that says "Do not display this alert on the homepage".  After that, every page started generating a cache entry in the cache render table.  Example cache key: ```entity_view:block:localgov_alert_banner:[languages:language_interface]=en:[theme]=localgov_theme:[url.path]=/1524:[user.permissions]=71...```

The ```url.path``` cache context may be coming from the condition_field module.  In which case we probably can't do much.  

In contrast, when I changed the visibility rule to only display that alert on the homepage, there were only two cache keys in the cache_render table irrespective of how many pages I visited.  That's good news :)

Given all this, I think I can live with PR.  We are not even using any Visibility rule in our site given that this is so new :)

Adnan-cds:
By the way, I noticed something very odd about the Visibility rule.  It is showing up as part of the Alert banner!  I have attached a screenshot:
screenshot-alert-banner-with-visibility-rule

Possibly caused by [this piece of config](https://github.com/localgovdrupal/localgov_alert_banner/blob/04a574780aff34611dbd332631693b49b61c5f4f/config/install/core.entity_view_display.localgov_alert_banner.localgov_alert_banner.default.yml#L22)

andybroomfield:
@Adnan-cds I noticed that and thought I upgraded wrong so removed it in managed display.
I'd start a new PR that moves that config to disabled

Adnan-cds:
> I'd start a new PR that moves that config to disabled

Thanks Andy :)

src/Entity/AlertBannerEntity.php
Adnan-cds: I actually preferred ```Cache::mergeContexts(parent::getCacheContexts(), $contexts);```.  That seem to be what most modules do.  [Example from the book module](https://git.drupalcode.org/project/drupal/-/blob/9.2.x/core/modules/book/src/Plugin/Block/BookNavigationBlock.php#L186).  Not a big deal though.

andybroomfield: I've pushed changes with moving them to the consturcter...
However is as @Adnan-cds says, if the standard Drupal way is to use the `Cache::mergeContexts(parent::getCacheContexts(), $contexts);` from `getCacheContexts()` then I will revert it to that.

andybroomfield:
Have now updated this branch.
Following @Adnan-cds suggestion, we're using the method that Drupal core and other modules use, which is to merge the context provided by the alert banner entity with that provided by the parent::getCacheContexts. This should still allow other modules to add custom cache contexts and tags as they will still be retrived from entityBase.php

src/Entity/AlertBannerEntity.php
Adnan-cds: Looks good to me.  Thanks.

> It may be better to actually move the logic here out of the getCacheContexts method and have it run by the constructor to ensure it's only run once.

I am not ignoring Steve's concern.  If it later turns out that this is indeed slowing down things then we can also consider using the good old [drupal_static()](https://api.drupal.org/api/drupal/core%21includes%21bootstrap.inc/function/drupal_static/9.1.x).

Adnan-cds (approved):
I am happy with what I am seeing.

andybroomfield:
@stephen-cox are you happy with this change also, as its your request that is blocking?

src/Entity/AlertBannerEntity.php
stephen-cox: I think this is fine. I don't fully understand why core doesn't make more use of the `addCacheContexts` property, but happy to go with this.

stephen-cox (approved):

PR #155, opened 2021-07-02
Fix adding cache contexts only when it has an effect on visibility

### LINKED ISSUE #154: Cache entries being added for alert banners that are not included on a page
From @Adnan-cds 
> Hi Andy,
I have finished my testing. It was doing great until I added a Visibility condition that says "Do not display this alert on the homepage". After that, every page started generating a cache entry in the cache render table. Example cache key: entity_view:block:localgov_alert_banner:[languages:language_interface]=en:[theme]=localgov_theme:[url.path]=/1524:[user.permissions]=71...
> 
> The url.path cache context may be coming from the condition_field module. In which case we probably can't do much.
> 
> In contrast, when I changed the visibility rule to only display that alert on the homepage, there were only two cache keys in the cache_render table irrespective of how many pages I visited. That's good news :)
> 
> Given all this, I think I can live with PR. We are not even using any Visibility rule in our site given that this is so new :)

andybroomfield:
Fix #154

eg. When a banner is set to show on specific pages, only add the cache context for url.path
when the current path is one of those pages.
When a banner is set to be hidden on specific pages, only add the cache context for url.path
when the current path is one of those pages, and skip on the rest.

### DIFF
--- a/src/Entity/AlertBannerEntity.php
+++ b/src/Entity/AlertBannerEntity.php
@@ -398,6 +398,18 @@ class AlertBannerEntity extends EditorialContentEntityBase implements AlertBanne
       $conditions_config = $this->get('visibility')->getValue()[0]['conditions'];
 
       foreach ($conditions_config as $condition_id => $values) {
+
+        // Skip adding contexts for banners that arn't visible.
+        // @see #154
+        // Show banner on restricted pages, skip when hidden.
+        if (!$this->isVisible() && $values['negate'] == 0) {
+          continue;
+        }
+        // Hide banner on restricted pages, skip when visible.
+        if ($this->isVisible() && $values['negate'] == 1) {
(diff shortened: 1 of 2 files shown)


andybroomfield:
This is a super initial fix @Adnan-cds however let me know if it works for you.

finnlewis:
@ekes @stephen-cox thoughts on this?

Adnan-cds:
I will test this and get back to you in a day or so.  This fix is not urgent (at least for me), so no need to hold the release for this.

Adnan-cds (approved):
Works for me.  Thanks :)

finnlewis:
Thanks @Adnan-cds 

@andybroomfield can you go ahead and merge this?

Just trying to jolly as many pull requests along by the end of the sprint!

PR #158, opened 2021-07-08
Cache clear can now rely on cache tag invalidation

### LINKED ISSUE #126: Is it nessecary to clear the cache when publishing a banner? (Improve cache tags / contexts)
There is quite a performance slump when publishing a new alert banner, as the cache needs to be cleared.
Lets improve the cache tags / contexts as we should be able to rely on those for banner publication.

Adnan-cds:
This comment in the dropped code says it best:
> Better to use cache tags instead of doing a full flush?"

#resolves #126

### DIFF
--- a/src/Entity/AlertBannerEntity.php
+++ b/src/Entity/AlertBannerEntity.php
@@ -162,8 +162,10 @@ class AlertBannerEntity extends EditorialContentEntityBase implements AlertBanne
   public function postSave(EntityStorageInterface $storage, $update = TRUE) {
     parent::postSave($storage, $update);
 
-    // Better to use cache tags instead of doing a full flush?
-    drupal_flush_all_caches();
+    // Needed by the entity.localgov_alert_banner.status_form route to function
+    // correctly during *test runs*.  Not the ideal solution, but less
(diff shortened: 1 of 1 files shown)


andybroomfield:
> 1) Drupal\Tests\localgov_alert_banner\Functional\AlertConfirmationTest::testAlertConfirmation
The link Remove banner was not found on the page.

Failing consistently so not a Github tests issue. This implies the confirmation form for removing a banner is not cache invalidating properley.

Adnan-cds:
Hi Andy,
The "Put banner live/Remove banner" route is using an [outdated copy of the Alert banner entity](https://github.com/localgovdrupal/localgov_alert_banner/blob/a0f2c7dfee80c8f393ac3327d86fce8f85571a79/src/Controller/AlertBannerEntityController.php#L221) during the test run.  I couldn't pin point the exact cause.  I have replaced drupal_flush_all_caches() with the less destructive [router rebuild](https://git.drupalcode.org/project/drupal/-/blob/9.2.x/core/lib/Drupal/Core/Routing/RouteBuilderInterface.php#L43).

andybroomfield (approved):
Ok, this works for me.

PR #161, opened 2021-07-20
Permissions to view banner and the entity page fixes

andybroomfield:
Adds view all and seperate view entity page permissions
Adds default view all to the anon and authenticated user role
Re-adds removed routesubscirber and access subcriber taken out in #135

### DIFF
--- a/config/install/user.role.emergency_publisher.yml
+++ b/config/install/user.role.emergency_publisher.yml
@@ -15,3 +15,4 @@ permissions:
   - 'use localgov_alert_banners transition publish'
   - 'use localgov_alert_banners transition unpublish'
   - 'view all localgov alert banner entities'
+  - 'view all localgov alert banner entity pages'
--- a/localgov_alert_banner.install
+++ b/localgov_alert_banner.install
@@ -204,3 +204,11 @@ EOY;
     user_role_grant_permissions('emergency_publisher', $permissions);
   }
 }
+
+/**
+ * Add view alert banner permission to anonymous and authenticated users.
+ */
+function localgov_alert_banner_update_9002() {
+  user_role_grant_permissions(RoleInterface::ANONYMOUS_ID, ['view all localgov alert banner entities']);
+  user_role_grant_permissions(RoleInterface::AUTHENTICATED_ID, ['view all localgov alert banner entities']);
+}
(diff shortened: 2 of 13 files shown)


andybroomfield:
~~I'm going to remove the permissions generator here and only have global permissions for all banner types. 
Is that ok with everyone?~~

andybroomfield:
Moving this from draft to review.

This adds seperate permissions for viewing an alert banner, and accessing the alert banner entity page (for preview / revisions).

An update / install hook will grant anonymous and authenticated user permission to view all alert banners, however they will not be /admin/content/alert-banner/{id} or /admin/content/alert-banner/{id}/revisions/{revision_id}/view.
Only those with the emergency publisher role should be able to access this path (restored the old route subscriber).

Additional permissions for per alert banner type are still included, this is to cover a scenario whereby a banner would be intended for internal use only (notifiing communications team of upcoming changes)  which would be a seperate banner type. In such a case the permissions for each banner type would need to be changed by site builders to grant the appropritate access. 
(Above could also be handled by adding a user role condition to the banner, but that isn't enabled by default on our condition field for banners).

Tests where already testing access to banners and routes, but no access control on the banners themselves meant they always passed (anon users never had view permission, so they couldnt access the pages, but we also never checked if they could view the banner when displayed in a block. Now that we do check we had to give them permission to view the banner but still stop them accessing the route). Thats the happy path tests working, should really add a test where a user does not have access to a banner and check it can't be viewed (manual testing shows this is the case).

Test fails are due to other Localgov Drupal modules.

finnlewis:
Discussing on our Merge Monday call, @ekes would like a trivial test for this to confirm that the correct user can see it or not, (and not cached).

andybroomfield:
Finally looks like I have fixed this, so it is now ready for review again.

ekes (approved):
\o/ tests

finnlewis:
Just discussing this at merge Monday meeting.

Noting that the tests are failing on the 1.x branch, but this is not a blocker to merging this as we would like to remove support for the 1.x branch.



@andybroomfield  will create a separate branch to remove the testing against 1.x

https://github.com/localgovdrupal/localgov_alert_banner/blob/1.x/.github/workflows/test.yml#L21-L23

Otherwise we're happy to merge this now that this has the tests to confirm the permissions are working.

PR #167, opened 2021-09-01
Fix Github CI tests

stephen-cox:


finnlewis (approved):
Tests are working, looks good to me!

PR #172, opened 2021-09-07
Add routing for accessing revision paths

### LINKED ISSUE #171: Viewing a revision results in 403 error
When looking at revisions, its not actully possible to view a revision.
This is even with user 1 account.

Steps to reproduce:
1. on an alert banner entity, go to revisions tab
2. click a revision that is not the latest
3. witness 403 access denied error

What should happen:
user is able to view revision.

This was found while checking #169 but I think the error is wider.
Localgov Workflows was installed at the time.

stephen-cox:


andybroomfield:
Viewing now works for me, assume its same process for revert and delete routes?

Screenshot 2021-09-08 at 8 59 48 am

stephen-cox:
>  assume its same process for revert and delete routes?

I have tried this and it's not working for these routes. Can't see anything in the docs to say why it doesn't: https://www.drupal.org/docs/8/api/routing-system/parameters-in-routes/parameter-upcasting-in-routes

I'm off for a little while now. Someone else is welcome to pick this up before I return. Otherwise will take a look in a couple of weeks.

finnlewis:
@stephen-cox is now away, is this something @andybroomfield you can take over?

andybroomfield:
This fixes the tests, including the additional permissions tests I have added.
This should be ready now, but someone else needs to review (check that the /revisions/revert and /revision/delete pages are accessbile.

It's a little odd that the revert form patttern is now different.
eg.
````php
  /**
   * {@inheritdoc}
   */
  public function buildForm(array $form, FormStateInterface $form_state, AlertBannerEntityInterface $localgov_alert_banner_revision = NULL) {
    $this->revision = $localgov_alert_banner_revision;
    $form = parent::buildForm($form, $form_state);

    return $form;
  }
````
whereas on nodes it is
````php
  /**
   * {@inheritdoc}
   */
  public function buildForm(array $form, FormStateInterface $form_state, $node_revision = NULL) {
    $this->revision = $this->nodeStorage->loadRevision($node_revision);
    $form = parent::buildForm($form, $form_state);

    return $form;
  }
````

I assume it's due to this issue https://www.drupal.org/project/drupal/issues/2730631 and not anything that is going to break in future, but a second pair of eyes over then change would be helpful.

andybroomfield:
Tests against profile 1.x (Drupal 8) are failing due to directories / events, localgov_geo and roles helper.

finnlewis:
Thanks @andybroomfield 

@ekes and @andybroomfield - might this mean we need different 1.x and 2.x branches for those projects?

finnlewis:
...or do we split localgov_alert_banner into a 1.x branch and a 2.x branch? 

Or just remove active support / testing for localgov 1.x from localgov_alert_banner? 

We'll need to remove these lines: https://github.com/localgovdrupal/localgov_alert_banner/blob/1.x/.github/workflows/test.yml#L21-L23

finnlewis (commented):
The functionality all works for me locally. I can view and delete revisions. The code looks logical enough and coding standards pass. Just need to decide what to do about the failing 1.x tests on localgov_event and localgov_directories.

ekes:
1.x events and directories shouldn't be downloading 1.1.x geo; they should have their version at 1.0.x
It's only the test that fails - so they could run 1.1.x - but... so if they were set so composer could download 1.x.x it should be more specific?

finnlewis:
Thanks @ekes - I think I understand, testing this our with a PR on the https://github.com/localgovdrupal/localgov_events/pull/46/files

finnlewis (approved):
Test pass! Let's merge this!

PR #178, opened 2021-10-05
Add alert banner title to the hide button aria-label attribute

### LINKED ISSUE #177: Alert banner hide button aria-labels lack context - all 'Hide alert'
[Accessibility audit ref 62](https://docs.google.com/spreadsheets/d/1t8rRIQe_1kogxPBZDKcsbf8YxtEfC9iGJvidkagNSXY/edit#gid=1779042480&range=D63)

The aria-label for the hide button on alert banners is set to 'Hide alert'. If there are multiple alert banners live at the same time, they all use the same aria-label, so there's no context for the user to know which alert the button will hide.

Suggest that the aria-label attribute is extended to include the alert title, eg: "Hide alert: Shakespeare is dead"

danchamp:
Closes #177 

Appends alert banner title to the 'Hide alert' button aria-label.

There may be a better way of accessing the plain text of the title?

### DIFF
--- a/templates/localgov-alert-banner.html.twig
+++ b/templates/localgov-alert-banner.html.twig
@@ -44,7 +44,7 @@
             {{ content.title }}
           </h2>
         {% endif %}
-        
+
         <div class="localgov-alert-banner__body">
           {{ content|without('title', 'link') }}
         </div>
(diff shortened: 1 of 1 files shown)


markconroy:
Looks good to me.

PR #180, opened 2021-10-05
Update Github Actions config to use standard config

stephen-cox:
Part of localgovdrupal/.github#10


finnlewis (approved):
Looks good to me!

finnlewis (approved):
Looks good to me, nice one @stephen-cox !

PR #192, opened 2022-04-07
Update Github actions configuration to try to test against PHP 7.4 an…

finnlewis:
…d 8.1 and only run coding standards and phpstan checks against code in this repository.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -183,4 +183,4 @@ jobs:
           # environmental variables, so it's necessary to change the config file directly.
           sed -i "s#http://localgov.lndo.site#http://drupal#" ./html/phpunit.xml.dist
           docker exec -t drupal bash -c 'chown docker:docker -R /var/www/html'
-          docker exec -u docker -t drupal bash -c "cd /var/www/html && ./bin/paratest --processes=4"
+          docker exec -u docker -t drupal bash -c "cd /var/www/html && ./bin/paratest --processes=4 modules/contrib/${{ github.event.repository.name }}"
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -17,10 +17,9 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        include:
-          - localgov-version: "2.x"
-            drupal-version: "~9.1"
-            php-version: "7.4"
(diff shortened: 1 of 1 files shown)


finnlewis:
Hey people - how's this looking? 

I think I've configured the tests to try both PHP 7.4 and 8.1 and to only run coding standards, phpstan and phpunit tests on code from this repository.

It looks like the tests are passing too! 

If this looks like a good pattern we can copy this for Drupal modules and set up a fuller integration test for the profile.

Any comments or suggestions welcome!

Adnan-cds (approved):
Looks good actually :)

andybroomfield (approved):
Looks good and they run ok.
Little concerned there is a skipped test... Do we know what it is?

finnlewis:
Not sure that the skipped test is @andybroomfield .

Here's the output of my local test run, which does have some deprecation notices, but does not appear to skip any of the 16 tests: 

https://github.com/localgovdrupal/localgov_alert_banner/issues/193

andybroomfield:
@finnlewis it is on the [checks results page](https://github.com/localgovdrupal/localgov_alert_banner/runs/5872911846?check_suite_focus=true)
````
Run mkdir -p ./html/web/sites/simpletest && chmod 777 ./html/web/sites/simpletest
Running phpunit in 4 processes with /var/www/html/vendor/phpunit/phpunit/phpunit

Configuration read from /var/www/html/phpunit.xml.dist

...............S                                                  16 / 16 (100%)

Time: 01:31.301, Memory: 10.00 MB

OK, but incomplete, skipped, or risky tests!
Tests: 16, Assertions: 23[9](https://github.com/localgovdrupal/localgov_alert_banner/runs/5872911846?check_suite_focus=true#step:5:9), Skipped: 1.
````

finnlewis:
Hey @andybroomfield , yes I see the output from the Github checks, but I don't know what the skipped test is as it does not do the same locally. The 16th test appears to be : 

Scheduling (Drupal\Tests\localgov_alert_banner\Kernel\Scheduling)
 ✔ Alert banner scheduling

Any idea why that would skip in some circumstances? 

I guess it is this one: 

https://github.com/localgovdrupal/localgov_alert_banner/blob/release-1.x/tests/src/Kernel/SchedulingTest.php

PR #194, opened 2022-04-11
Default the unpublish other alert banners to off

### LINKED ISSUE #160: Default the unplublish other alert banners to off
... This is now the default behaviour at BHCC (done with a hook form alter).
I think it should actully be default behaviour now as this dates from a time when converting from only 1 banner to allowing multiple where multiple was a rare use case. With conditional banners now added this can be extra confusing.

I'd actully go one step further and remove this feature now, as we can't assume all banners should really be unpublished.

andybroomfield:
Fix #160.

This matches behaviour at BHCC.


finnlewis (approved):
Happy with this. Add a change record to the release notes!

PR #195, opened 2022-04-11
Set weight of alert banner block to -100

### LINKED ISSUE #153: Alert banners appear below Subsite banners
Subsite banners have a weight in config of -7, meaning they appear above alert banners in the Banner region:

https://github.com/localgovdrupal/localgov_subsites/blob/2.x/config/optional/block.block.subsitebanner.yml

Screenshot 2021-07-01 at 18 24 25

Propose setting Alert Banners default weight to -100 or similar.

andybroomfield:
To fix #153 .


danchamp (approved):
Looks good, thanks @andybroomfield

finnlewis (approved):
Reviewing at Merge Monday - all happy with this

PR #196, opened 2022-04-11
Remove role="alert"

### LINKED ISSUE #146: Should we remove role="alert" from alert banners
When testing a page in AXE, I am told that `role="alert"` is not valid on the article element.

Two things we can do:
1. Remove the `article` element and place a `div` there instead, or
2. remove the `role="alert"` attribute.

I'm open to debate on this one, but I _think_ role="alert" may not be correct here. I think that type alert is for dynamic content, such as form validation if someone hasn't filled out a form properly for example. 

Here's the MDN notes on it:
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_alert_role

and some notes on live regions
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions

andybroomfield:
Fix #146.

In case we wish to pick up discussion and proceed.

### DIFF
--- a/templates/localgov-alert-banner.html.twig
+++ b/templates/localgov-alert-banner.html.twig
@@ -32,7 +32,7 @@
   ]
 %}
 
-<article {{ attributes.addClass(classes) }} role="alert">
+<article {{ attributes.addClass(classes) }}>
   {# Begin Wrapper #}
   <div class="localgov-alert-banner__wrapper">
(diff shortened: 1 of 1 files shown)


markconroy (approved):
Looks good to me.

PR #200, opened 2022-04-19
Replace use of jquery cookie with js-cookie

### LINKED ISSUE #197: Replace jQuery.cookie with core/js-cookie
As jQuery cookie is removed in Drupal 10.
The library is a dependency (and used by) js/alert_banner.js.
See https://www.drupal.org/node/3104677

andybroomfield:
Fix #197

This is just switching the library and supporting code to address
Drupal 10 deprecation of jquery.cookie

### DIFF
--- a/js/alert_banner.js
+++ b/js/alert_banner.js
@@ -5,19 +5,19 @@
  * This is so if the alert changes, the banner is reshown.
  */
 
-(function($, Drupal, drupalSettings) {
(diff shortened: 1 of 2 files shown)


andybroomfield:
Have manual checked this, I suspect though that the dismissing the alert banner is the skipped test.

finnlewis (approved):
looks good to me @andybroomfield and tests pass.

PR #204, opened 2022-04-27
Add alert banner workflow as role dependency

### LINKED ISSUE #203: Fix non-existent permissions on Drupal 10
> RuntimeException: Adding non-existent permissions to a role is not allowed. The incorrect permissions are "use localgov_alert_banners transition publish", "use localgov_alert_banners transition unpublish". in Drupal\user\Entity\Role->calculateDependencies() (line 206 of core/modules/user/src/Entity/Role.php).

This is when installing localgov alert banner on Drupal 10 alpha 3, but not installing scheduled transitions.

andybroomfield:
Fix #203 

Add workflows.workflow.localgov_alert_banners config as a dependency to user.role.emergency_publisher role.
This is required to avoid error when installing the module under Drupal 10. 
It makes sure the alert banner workflow config is installed before installing the emergency publisher role.

### DIFF
--- a/config/install/user.role.emergency_publisher.yml
+++ b/config/install/user.role.emergency_publisher.yml
@@ -4,6 +4,8 @@ dependencies:
   enforced:
     module:
       - localgov_alert_banner
+  config:
+   - workflows.workflow.localgov_alert_banners
 id: emergency_publisher
 label: 'Emergency publisher'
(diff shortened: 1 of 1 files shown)


stephen-cox (approved):
Yep - makes sense. Works fine on D9.

PR #206, opened 2022-04-27
Hide the visibility conditions on display

### LINKED ISSUE #205: Condition field is being displayed by default
On an initial install, the condition field plugin is displaying its output by default on the alert banners.

Screenshot 2022-04-27 at 11 35 11 pm

andybroomfield:
Fix #205 

Add the visibility conditions to the hidden section of the manage display config.


stephen-cox (approved):
Looks to only effect Drupal 10 as Drupal 9 doesn't display the visibility field. Only tested on D9. Works fine.

PR #212, opened 2022-05-31
Don't overwrite Scheduled Transitions bundle configuration

### LINKED ISSUE #211: Enabling scheduled transitions overwrites previous bundle configuration
Part of https://github.com/localgovdrupal/localgov_workflows/issues/25

The code that configures Scheduled Transitions when it's enabled overwrites previous bundle configuration. One manifestation of this is it breaks the Review Date widget.

stephen-cox:
Fixes #211


andybroomfield (approved):
Tested by enabling localgov_workflows and then localgov_alert_banners and the editoral workflow was still intact with the relevant content types.

PR #217, opened 2022-06-07
Officially support Drupal 10

andybroomfield:
Mark module as supporting Drupal 10, now that condition field also 
supports Drupal 10 and field group is no longer a dependancy.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -6,8 +6,7 @@
     "minimum-stability": "dev",
     "license": "GPL-2.0-or-later",
     "require": {
-        "drupal/condition_field": "^2.0",
-        "drupal/field_group": "~3.1"
+        "drupal/condition_field": "^2.0"
     },
     "require-dev": {
         "drupal/scheduled_transitions": "^2.1"
--- a/localgov_alert_banner.info.yml
+++ b/localgov_alert_banner.info.yml
@@ -7,7 +7,6 @@ dependencies:
   - drupal:block
   - drupal:content_moderation
   - drupal:field
-  - drupal:field_group
   - drupal:link
   - drupal:node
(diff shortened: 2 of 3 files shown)


finnlewis:
@Adnan-cds we're hoping you could take a look at this with respect to the full page sub-module.

Adnan-cds:
Looks okay to me.  The [localgov_alert_banner_full_page submodule explicitly says](https://github.com/localgovdrupal/localgov_alert_banner/blob/1.x/modules/localgov_alert_banner_full_page/localgov_alert_banner_full_page.info.yml) that it is compatible with Drupal 8 and 9 only.  So there shouldn't be any confusion once the main module starts supporting Drupal 10.

andybroomfield:
@Adnan-cds Just to confirm you would be ok for me to make a release announcing support for Drupal 10 without updating the sub module?
Do we know what needs to be updated before releasing? Do we need the field group dependency?

Adnan-cds:
> Do we need the field group dependency?

Hi Andy, I just tested the submodule in a fresh site without the field_group module and it works fine.  So if you can please [remove the dependency on the field_group module](https://github.com/localgovdrupal/localgov_alert_banner/blob/9ee9c9bd2c872ca804f50e910c0d68e244ed4810/modules/localgov_alert_banner_full_page/localgov_alert_banner_full_page.info.yml#L9), that would be great :)

finnlewis:
Discussing in Merge Monday, @andybroomfield will remove the dependency on drupal:field_group and mark as supported for Drupal 10.

Adnan-cds (approved):
Looks good to me.  Thanks :)

Issue #218, opened 2022-06-24
Support new Drupal manage permissions tab


andybroomfield:
[See article](https://www.droptica.com/blog/whats-new-drupal-94-review-latest-version/) and [Drupal documentation](https://www.drupal.org/node/3242827).

We should support the manage permissions form. It requires making sure permissions are assigned to an entity. This can be done using the new [BundlePermissionHandlerTrait](https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Entity%21BundlePermissionHandlerTrait.php/trait/BundlePermissionHandlerTrait/9.3.x) but this only exists in Drupal 9.3. We could either copy the trait into the permission callback to keep compatibility, or we could start a 2.x branch now and start supporting new Drupal features.

andybroomfield:
Done!

PR #219, opened 2022-06-24
Support for new entity permissions form

andybroomfield:
Ref #218

Add new annotation to get the alert banner permissions form on the manage alert banner type
entity.
Also changes the alert banner permissions generator to make sure are assigned to the alert
banner entity using BundlePermissionHandlerTrait

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -6,7 +6,8 @@
     "minimum-stability": "dev",
     "license": "GPL-2.0-or-later",
     "require": {
-        "drupal/condition_field": "^2.0"
+        "drupal/condition_field": "^2.0",
+        "drupal/core": "^9.4 || ^10.0"
     },
     "require-dev": {
         "drupal/scheduled_transitions": "^2.1"
--- a/localgov_alert_banner.info.yml
+++ b/localgov_alert_banner.info.yml
@@ -1,7 +1,7 @@
 name: 'LocalGov Alert Banner'
 type: module
 description: 'Provides a sitewide alert banner for urgent alerts.'
-core_version_requirement: ^8.8 || ^9 || ^10
+core_version_requirement: ^9.4 || ^10
 package: LocalGov Drupal
 dependencies:
   - drupal:block
--- a/localgov_alert_banner.permissions.yml
+++ b/localgov_alert_banner.permissions.yml
@@ -23,4 +23,4 @@ manage all localgov alert banner entities:
 
 # Per bundle permissions.
 permission_callbacks:
-  - \Drupal\localgov_alert_banner\AlertBannerEntityPermissions::generatePermissions
+  - \Drupal\localgov_alert_banner\AlertBannerEntityPermissions::alertBannerTypePermissions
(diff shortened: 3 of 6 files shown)


stephen-cox:
Discussed at Merge Monday. Thinking was we can add a dependency to Drupal 9.4 and release it as 1.5.x

finnlewis:
Discussed in Merge Monday, @andybroomfield will reawaken this now that 9.4 is out and 9.3 is deprecated.

finnlewis:
Still on Andy's backlog, probably not get to this for another month at least.

andybroomfield:
Updated this PR dropping support for Drupal 8 and everything below Drupal 9.4 (core 9.4+ or 10).
This means adding drupal/core to composer.json so on the off chance there are sites still below 9.4 requiring this, they wont get the update.

Once #244 and #246 are merged, they will make up the final release of the 1.5.x series.
This will then form the first release of 1.6.x

andybroomfield:
This is ready for review now.
You'll now get a new manage permissions tab when editing an alert banner type under structure -> alert banners -> alert banner.

/admin/structure/alert-banner-types/localgov_alert_banner/permissions

Screenshot 2023-05-15 at 9 41 23 pm
  

Note: Emergency publisher doesn't need the manage permission, as they have the manage all alert banners permission.
Screenshot 2023-05-15 at 9 40 57 pm

stephen-cox (commented):
This adds the permissions form, but I'm not sure if it's working as expected. If it is working it's a little confusing.

If I remove the 'View alert banner entities of type Alert banner' permission on the new permissions tab it has no effect because it's overridden by the global permission with the same name. I would expect the permission on a banner type to override the global permission when removing permissions.

Adnan-cds:
> If it is working it's a little confusing.

I think it is working.  But it is confusing indeed :D

Adnan-cds:
Code looks good to me.  But we really need to be sure if this is working or not.

andybroomfield:
Yes, thats expected behaviour (global permission overrides per entity type), as the view all are assigned to anon user and auth user, and manage to emergrency publisher.
I did try and see if it was possible to add some explinationary copy about this, but it doesn't seem possible to target this specific instance of the permissions form.

We might consider actully removing the global permissions, that would require update hook to re-assign them.
I note that we are in fact [adding permissions on the entity type when created](https://github.com/localgovdrupal/localgov_alert_banner/blob/3077fe859f4db2a10827081db62d6236652bd4a5/src/Entity/AlertBannerEntityType.php#L106) so it wouldn't be much to extend that for the manage permissions.

My only counter to this is that some teams may have set up a different role and granted a global permission.

finnlewis:
Could do with testing to confirm that the sub-set of permissions page works as expected. 

I'm happy to test further! :)

andybroomfield:
Would like to pick this up again.
Are we ok with a plan of:
- Merge this so the new permissions tab is present (Support a new Drupal feature)
- Do some discovery on the permission model, potentially removing the global permissions

We could then look at supporting the new permissions tab on other custom entities such as geo.

finnlewis:
@stephen-cox also happy with this approach... so I'll let you confirm before / if testing further.

stephen-cox (approved):
This is fine. Happy to merge this and rationalise the permissions in another ticket.

PR #221, opened 2022-09-08
Fix: Cookie handling

Adnan-cds:
Replaced core/jquery.cookie with core/js-cookie.  As well as being deprecated, core/jquery.cookie also broke our cookie handling in the Full page alert banner.

@see https://www.drupal.org/node/3104677


andybroomfield (approved):
Tested and works.
Shall we make this a release @Adnan-cds

PR #222, opened 2022-09-08
1.5.1

andybroomfield:
Fix cookie handling with full page alert banner.


Adnan-cds (approved):
Thanks a lot to Andy.

PR #225, opened 2022-10-24
Set the title colour for notable person messages

opdavies:
When using the "Death of a notable person" alert type which sets a black background colour, the title - which is also set to black - isn't visible.

This PR sets an explicit text colour for the title within that alert type - ensuring that it can be seen and read.

I did consider opening this PR against the `localgov_base` theme but thought that it made more sense to do so within the alert banner module, although I'd be happy to re-submit it if needed.

## UI changes

Before:

Screenshot 2022-10-24 at 01-12-20 Welcome! test

After:

Screenshot 2022-10-24 at 01-18-37 Welcome! test

### DIFF
--- a/css/localgov-alert-banner.css
+++ b/css/localgov-alert-banner.css
@@ -60,6 +60,9 @@
 .localgov-alert-banner--notable-person {
   background-color: #000;
 }
+.localgov-alert-banner--notable-person .localgov-alert-banner__title {                                                                                        │
+  color: #fff;
+}
 
 .localgov-alert-banner__wrapper {
   max-width: 73.125rem;
--- a/css/localgov-alert-banner.css
+++ b/css/localgov-alert-banner.css
@@ -13,6 +13,11 @@
   color: #fefefe;
 }
 
+/* Reset the title colour as some themes assign a colour to h2 */
+.localgov-alert-banner .localgov-alert-banner__title {
+  color: inherit;
+}
+
 .localgov-alert-banner__close {
   padding: .5rem .9375rem;
   line-height: 1.5;
@@ -59,9 +64,6 @@
 /* Death of a notable person */
 .localgov-alert-banner--notable-person {
   background-color: #000;
-}
-.localgov-alert-banner--notable-person .localgov-alert-banner__title {                                                                                        │
-  color: #fff;
 }
 
 .localgov-alert-banner__wrapper {


finnlewis:
Hey @opdavies, thanks for this. 

Looks good to me, but hopefully @andybroomfield can take a look and we can merge and release at our "Merge Monday" call today.

finnlewis:
@markconroy @msayoung how's this look for you?

finnlewis (approved):
Discussing at Merge Monday, looks good to @markconroy and happy to merge it. 

Might just hang on to check with @andybroomfield as the lead maintainer.

andybroomfield:
Thanks @opdavies 
I belive this change is best made in localgov_base, as that is where the title colour is being set.
(The h2 colour sets it).
Although looking on a default install, even base Olivero theme is overriding the h2 colours.
I suppose its how we feel about having it here, as it will then need to be overridden for all theme builders.

Maybe 
````css
.localgov-alert-banner .localgov-alert-banner__title {
  color: inherit;
}
````
Instead?
Would we need to apply it to the other banners and fields also?

**Comment by markconroy** [comment]
Here's a PR for the `localgov_base` theme: https://github.com/localgovdrupal/localgov_base/pull/311

@andybroomfield I don't think it will need to be overridden for most themers, as I guess most themes will be happy with a black background and white text for this particular alert.

File: css/localgov-alert-banner.css [hunk 1]
```diff
@@ -60,6 +60,9 @@
 .localgov-alert-banner--notable-person {
   background-color: #000;
 }
+.localgov-alert-banner--notable-person .localgov-alert-banner__title {                                                                                        │
+  color: #fff;
+}
```
andybroomfield: I don't think we should have the `|` there.
Lets change this to 
````css
.localgov-alert-banner .localgov-alert-banner__title {
  color: inherit;
}
````
This is so it applies a fix to all banners, and that they use the colour assigned to it via `.localgov-alert-banner`. and can then be overridden in themes by just appling a color to that class.

andybroomfield (dismissed):
Thanks for the fix @opdavies.
See my comment which I think would be the way to fix this.

finnlewis:
@opdavies any thoughts on the above?

andybroomfield:
Looks like I can push directly so maybe we can look at this on merge monday.

andybroomfield:
Do we need the Github test update for javascript tests?

finnlewis (approved):
Yay! Thanks @andybroomfield

PR #240, opened 2023-03-03
Update alert banner cookie to set samesite="lax"

rgcarr:
Added SameSite default setting to Cookie

### DIFF
--- a/js/alert_banner.js
+++ b/js/alert_banner.js
@@ -12,7 +12,7 @@
   function setAlertBannerHideCookie(cookie_tokens, token) {
     cookie_tokens.push(token);
     var new_cookie = cookie_tokens.join('+')
-    cookies.set('hide-alert-banner-token', new_cookie, { path: '/', expires: 30 });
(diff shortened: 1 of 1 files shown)


andybroomfield (approved):
Thanks for this @rgcarr 

I've tested and the banner still hides, and I'm no longer seeing a warning in firefox consiole.
We should review at some point if 'lax' is the right value here, and if secure (which EU cookie complicance uses) would be better. Also need to look at if we should send secure if under https. These can be looked at as seperate issues though.

andybroomfield:
Some more info on samesite
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite#browser_compatibility

rgcarr:
I did some testing with SameSite: None (which seemed most flexible option), but you have to implement Secure... and didn't think we can assume that a site is always running HTTPS (errors if Cookie is "Secure' but running HTTP). Lax seemed to be path of least resistance.

ekes:
And on the same day https://www.drupal.org/node/3275352 "Drupal now defaults to "Lax" for the SameSite session cookie attribute, and checks for valid values. Introduced in version: 
10.1.0" :smile:

Issue #245, opened 2023-05-11
Alert banner with empty Link text should show "More information", but nothing shows 


keelanfh:
On a clean install with localgov_base, there's this wording when creating an Alert banner ("If you don't write anything here, we will use: More information").

![image](https://github.com/localgovdrupal/localgov_alert_banner/assets/19519457/dd8e38e4-c41b-4a9f-be71-28cd8974b15e)

But looking at the banner on the site, I see no link at all:

![image](https://github.com/localgovdrupal/localgov_alert_banner/assets/19519457/ac055e44-6ff3-4933-93cc-e83f1c5daaf1)

andybroomfield:
With the default theme set to oliveo, the More information link is displayed.
Maybe this is an issue with Localgov base?

<img width="860" alt="Screenshot 2023-05-12 at 10 37 48 am" src="https://github.com/localgovdrupal/localgov_alert_banner/assets/1467480/30c22993-3bee-4235-b6a0-0461259b702f">

andybroomfield:
And a test on BHCC theme
<img width="1187" alt="Screenshot 2023-05-12 at 10 39 46 am" src="https://github.com/localgovdrupal/localgov_alert_banner/assets/1467480/f4ea4c8e-f4bf-41d5-bfe8-bd9ef8ee60fe">

andybroomfield:
This is set by [localgov_alert_banner.page.inc](https://github.com/localgovdrupal/localgov_alert_banner/blob/183c2d988de59b2d1c69a1d51644ac8d61a4246b/localgov_alert_banner.page.inc#L36)
````php
// Set default link text if field exists.
  if ($entity->hasField('link')) {
    if ($entity->get('link')->title === '') {
      $variables['content']['link'] = Link::fromTextAndUrl(t('More information'), Url::fromUri($entity->get('link')->uri))->toString();
    }
  }
````
Is localgov_base overriding this @markconroy ?

ctorgalson:
@andybroomfield sort of. It's that the existing assignment to `$variables['content']['link']` quoted above _changes_ the render array to from a LinkItem to a GeneratedLink, I believe.

@petrillek just noticed this too. I'll create a PR in a few minutes.

andybroomfield:
Closing as this looks fixed.

PR #249, opened 2023-05-16
Replaces link assignment when no custom text is provided

ctorgalson:
- The original code *changes* the type of the render array from LinkItem to GeneratedLink.
- This makes theming the entity slightly unpredictable.
- This change supplies the default text *without* changing the type of object provided to Twig.
- Re: #245

So this _does_ fix the issue, _and_ works as expected in e.g. `localgov_base`, but I don't know what to do for users who may have themed alert banner entities on the basis of the existing code. Feedback welcome :)

### DIFF
--- a/localgov_alert_banner.page.inc
+++ b/localgov_alert_banner.page.inc
@@ -35,8 +35,11 @@ function template_preprocess_localgov_alert_banner(array &$variables) {
 
   // Set default link text if field exists.
   if ($entity->hasField('link')) {
-    if ($entity->get('link')->title === '') {
-      $variables['content']['link'] = Link::fromTextAndUrl(t('More information'), Url::fromUri($entity->get('link')->uri))->toString();
+    $link = $entity->get('link');
+    if ($link->title === '') {
+      $link_value = $link->getValue();
+      $link_value[0]['title'] = t('More information');
+      $entity->set('link', $link_value);
     }
   }
 


andybroomfield:
I'm not seeing the more information anymore with this branch, just the bare link.

In localgov base

Screenshot 2023-05-22 at 11 35 22 am

and in Olivero

Screenshot 2023-05-22 at 11 31 58 am

localgov_alert_banner.page.inc
```diff
@@ -35,8 +35,11 @@ function template_preprocess_localgov_alert_banner(array &$variables) {
 
   // Set default link text if field exists.
   if ($entity->hasField('link')) {
-    if ($entity->get('link')->title === '') {
-      $variables['content']['link'] = Link::fromTextAndUrl(t('More information'), Url::fromUri($entity->get('link')->uri))->toString();
+    $link = $entity->get('link');
+    if ($link->title === '') {
+      $link_value = $link->getValue();
+      $link_value[0]['title'] = t('More information');
+      $entity->set('link', $link_value);
```
andybroomfield, line 42: I don't think this will work, as when altering entities, even right before display, we still have to do an entity->save() operation. But this would happen everytime it's displayed.

So we would either
- Move this to a preSave hook / event. This means it will add the text as the link text 'More information' and would not be changed if we ever updated the default
- Move this to a preprocess_link function, if such a function exists where the link text can be altered directly without saving the text.

I'm open to either, there are pros and cons to each, and it depends on how common it is for content designers to leave this blank.

I wonder if its possible to have default link text in Drupal core that is then overrideable.

andybroomfield (changes requested):
See above comments, it doesn't look like this change is having the intended effect.

localgov_alert_banner.page.inc
ctorgalson: > I don't think this will work, as when altering entities, even right before display, we still have to do an entity->save() 

This seems obvious when you say it :facepalm:.

On the other hand, It is definitely working for me right now with an unmodified `localgov_base` theme (authenticated, anonymous, any way I check). I'll have a look tomorrow and see if I can find out the reason (or the source of my hallucination...)

ctorgalson: > Move this to a preprocess_link function

Preprocess is an option too via `template_preprocess_field()`. With link fields when there's no `#title` provided `#url` and `#title` are identical, so it's possible to compare and override the `#title` value if they're the both set and equivalent.

finnlewis:
@ctorgalson just checking in on this in Merge Monday, @andybroomfield will have a look at your comments. Thanks!

localgov_alert_banner.page.inc
andybroomfield: I'd be tempted to go with a `template_preprocess_field()` in that case. Still showing as bare link when I try and on fresh installs.

andybroomfield:
Would like to try and see if we can revise this, maybe using the template_preprocess_field or hook_preprocess_field method?

finnlewis:
Discussing in Merge Monday. I'll test this to see if this is working for me. 
If not, maybe we rework following @andybroomfield suggestion above.
@ekes is wondering if the theme is doing this wrong. Themes should be getting the data from variables.  What is localgov_base doing here?

ekes:
Thinking about it. If I understand it. If the point is that it should have a value set for display in the theme then the theme should be getting the value from the theme values not the entity. However, if the intention is that the entity link value is supposed to return a default value that should be set at the entity level so not in a theme preprocess. Present patch in any case would remove setting the default in the theme variables.

ctorgalson:
Sorry I haven't had time to get to this: I have no contrib time at all right now.

@ekes 

> Thinking about it. If I understand it. If the point is that it should have a value set for display in the theme then the theme should be getting the value from the theme values not the entity.

If I understand what you're saying, the problem is caused _for_ the theme _by_ the module. As I mentioned at the top of the thread, the theme doesn't know whether to expect a `LinkItem` or a `GeneratedLink`. 

> The original code changes the type of the render array from LinkItem to GeneratedLink.

Looking at this PR, I think the logic could be altered to combine the original and the PR without bothering with preprocess: It could update the link, but instead of saving the entity etc, assign the current state of the variable to `$variables['content']['link']` like the original code does.

This would mean the theme gets the unmodified `LinkItem` when a title is set, _or_ a `LinkItem` with an added default title when it's not set.

finnlewis:
@alexburrows has offered to take a look. Thanks @alexburrows !

localgov_alert_banner.page.inc
alexburrows: After having a play around with what we have last night I concur with `template_preprocess_field()` route, i'll get the code sorted for this.

alexburrows: Welcome____Drush_Site-Install

```/**
 * Implements hook_preprocess_field().
 */
function localgov_alert_banner_preprocess_field(&$variables) {
  if ($variables['element']['#field_name'] == 'link' && $variables['element']['#bundle'] == 'localgov_alert_banner') {
    foreach ($variables['element']['#items'] as $delta => $item) {
      $link_item = $item->getValue();
      if (empty($item->getValue()['title'])) {
        $default_text = t('More information');
        $variables['items'][0]['content']['#title'] = $default_text;
      }
    }
  }
}
```

andybroomfield:
#276 replaces this.

PR #250, opened 2023-05-17
Feature: Group integration

Adnan-cds:
Provides:
- A Group plugin for Alert banners.
- A View for listing Alert banners belonging to Groups.
- User permissions to go along with the localgov_microsites_group module.

### DIFF
--- a/modules/group_alert_banner/README.md
+++ b/modules/group_alert_banner/README.md
@@ -1,0 +1,15 @@
+# Group Alert banner
+LocalGov Alert banner integration with the [Group module](https://www.drupal.org/project/group).
+
+## Configuration
+If you are using this module alongside the [localgov_microsites_group](https://github.com/localgovdrupal/localgov_microsites_group) module, then user permissions will be automatically setup during module installation.
+
+Otherwise enable at least the following permissions for **Groups**:
+- "Entity: View any alert banner entities" for both anonymous and authenticated users.
+- "Access Alert banner listing page" for group admins.
+
+## Banner listing
+Once this module is installed, an *Alert banner* tab should appear in each Group page.  This page will list all the banners belonging to a Group.
+
+## Todo
+- Automated tests.
--- a/modules/group_alert_banner/README.md
+++ b/modules/group_alert_banner/README.md
@@ -2,11 +2,15 @@
 LocalGov Alert banner integration with the [Group module](https://www.drupal.org/project/group).
 
 ## Configuration
+### User permissions
 If you are using this module alongside the [localgov_microsites_group](https://github.com/localgovdrupal/localgov_microsites_group) module, then user permissions will be automatically setup during module installation.
 
 Otherwise enable at least the following permissions for **Groups**:
 - "Entity: View any alert banner entities" for both anonymous and authenticated users.
 - "Access Alert banner listing page" for group admins.
+
+### Block placement
+If you are using the [localgov_microsites_base theme](https://github.com/localgovdrupal/localgov_microsites_base), which is the default theme of the [LocalGov Drupal Microsites distribution](https://github.com/localgovdrupal/localgov_microsites_project), an Alert banner block will be automatically placed in the *Header* region.  For all other themes, place an Alert banner block in a theme region towards the top of the page.
 
 ## Banner listing
 Once this module is installed, an *Alert banner* tab should appear in each Group page.  This page will list all the banners belonging to a Group.
--- a/modules/group_alert_banner/README.md
+++ b/modules/group_alert_banner/README.md
@@ -10,6 +10,3 @@ Otherwise enable at least the following permissions for **Groups**:
 
 ## Banner listing
 Once this module is installed, an *Alert banner* tab should appear in each Group page.  This page will list all the banners belonging to a Group.
-
-## Todo
-- Automated tests.
--- a/modules/group_alert_banner/group_alert_banner.group.permissions.yml
+++ b/modules/group_alert_banner/group_alert_banner.group.permissions.yml
@@ -1,0 +1,3 @@
+access localgov_alert_banner overview:
+  title: 'Access Alert banner listing page'
+  description: 'Access the Alert banner listing and management page, regardless of banner type.'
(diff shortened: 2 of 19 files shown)


Adnan-cds:
Mostly [lifted](https://git.drupalcode.org/project/group/-/blob/3.0.x/modules/gnode/src/Plugin/Group/Relation/GroupNode.php) from the gnode submodule of the group module.

modules/group_alert_banner/group_alert_banner.group.permissions.yml
```diff
@@ -0,0 +1,3 @@
+access localgov_alert_banner overview:
+  title: 'Access Alert banner listing page'
+  description: 'Access the Alert banner listing and management page, regardless of banner type.'
```
andybroomfield, line 3: This sounds like the kind of general permission we should have in the main module, how are we access controlling it at the moment? Why do we need the seperate permission?

andybroomfield:
As this is for microsites I'm happy to leave the review to those using it.
Be good to be able to have banners per group.
Raises a couple of questions on permissions, see comment above, though that might just be I don't have experince with group permissions.

modules/group_alert_banner/README.md
```diff
@@ -0,0 +1,12 @@
+# Group Alert banner
+LocalGov Alert banner integration with the [Group module](https://www.drupal.org/project/group).
+
+## Configuration
+If you are using this module alongside the [localgov_microsites_group](https://github.com/localgovdrupal/localgov_microsites_group) module, then user permissions will be automatically setup during module installation.
+
+Otherwise enable at least the following permissions for **Groups**:
+- "Entity: View any alert banner entities" for both anonymous and authenticated users.
```
andybroomfield, line 9: May have an issue with #219 and #256 if we remove this permission.

modules/group_alert_banner/group_alert_banner.module
```diff
@@ -0,0 +1,82 @@
...
+      ],
+      Drupal\localgov_microsites_group\RolesHelper::MICROSITES_EDITOR_ROLE => [
+        'access localgov alert banner listing page',
+        'use localgov_alert_banners transition create_new_draft',
+        'use localgov_alert_banners transition publish',
+        'use localgov_alert_banners transition unpublish',
+        'view all localgov alert banner entities',
+        'view all localgov alert banner entity pages',
+      ],
+    ],
```
andybroomfield, line 49: Granting this to the editors would be a major change in the permisions model, which is that alerts are by the emergency publisher. 
(Although I'm aware of councils that change this to also allow editors).
Is the emergency publisher role not something avalible on microsites?

modules/group_alert_banner/group_alert_banner.group.permissions.yml
andybroomfield: So I think I'm understanding that this is for the seperate group view, which is different from the main admin view. And we only want to grand group insiders the group admin view and not the main one?
(This is me trying to understand localgov microsites permissions)

andybroomfield:
My very quick run through within the localgov microsites distrubtion shows that this works well, thanks @Adnan-cds.
I think there are a few oddities, which could be a combination of microsites oddities, issues with domains, and understanding the group permissions vs global permissions. 

Permissions are something we need to look at, just to make sure we're not granting a bit more than we should. 
And we should redirect users after putting an alert banner live to the group version of the alert banner admin view.

I'm also noticing that when I click set live when creating a new alert banner, I'm taken to the view alert page, not the confirm form.

stephen-cox (changes requested):
This largely works fine. A couple of minor issues:

- As Andy points out, when adding alert the set live option doesn't work.
- When viewing the list of alert banners to the banner type is empty; maybe this isn't needed.

Just to note, that this branch is rather out of data and alert banners weren't displaying until I merged the latest changes in the 1.x branch.

If the add alert banner form redirects to the site live form when the set live field is selected then I think this is fine to merge.

Adnan-cds:
> And we should redirect users after putting an alert banner live to the group version of the alert banner admin view.

This is done now.

> when I click set live when creating a new alert banner, I'm taken to the view alert page, not the confirm form.

This is fixed as well.

The test runs are resulting in an [unrelated test failure](https://github.com/localgovdrupal/localgov_alert_banner/actions/runs/6074964785/job/16480308511?pr=250#step:5:21).  Is this a known issue?

modules/group_alert_banner/group_alert_banner.group.permissions.yml
Adnan-cds: > And we only want to grant group insiders the group admin view and not the main one?

That's right.

finnlewis:
@Adnan-cds would like @andybroomfield to review before merging. 

@andybroomfield would like @stephen-cox to review before merging! 😄

finnlewis:
@stephen-cox -  might you have time to review this one? 
@ekes  -  might you have time to review this one? 

@finnlewis can you test again?

and @millnut  - if you get bored?  😄

stephen-cox (approved):
Working well for me. Happy to approve.

One things to note is to get alert banners to display I had to add the alert banner block to the themes header region. I suggest adding optional block config for this to the Microsites Base theme as we have for other blocks: https://github.com/localgovdrupal/localgov_microsites_base/tree/2.x/config/install

Adnan-cds:
>  I suggest adding optional block config for this to the Microsites Base theme as we have for other blocks

That's a good point that I missed.  I remember having to manually add this block in our microsite instance.  How about adding it here in the group_alert_banner submodule as an optional block config?

stephen-cox:
> How about adding it here in the group_alert_banner submodule as an optional block config?

That would do the trick

Adnan-cds:
I have added an [optional config file](https://github.com/localgovdrupal/localgov_alert_banner/pull/250/files#diff-6303ce6fd5d4ee47488998ea4ab3a415a1e0058ea7afc8139b7750a4650ff11c) to add the Alert banner block in the localgov_microsites_base theme.  I have also explained the [block placement issue in the README](https://github.com/localgovdrupal/localgov_alert_banner/blob/feature/group-integration/modules/group_alert_banner/README.md#block-placement).  Please let me know if this will do.

stephen-cox:
Looks fine to me :+1:

Adnan-cds:
Great :) Merging then...

PR #255, opened 2023-05-23
[DO NOT MERGE] Fail the accesscheck

andybroomfield:
This should raise a deprecation warning!


andybroomfield:
So this should not have worked, as the query without accessCheck is deprecated and should have raised warnings.
Maybe because its a split query the rules wont find it.

PR #266, opened 2023-07-06
Fix coding standards issue (Use statements)

andybroomfield:
Fix use statements in the EntityPermissions callback


andybroomfield:
Think I need someone to clear test caches to force this one through?
@stephen-cox @finnlewis ?

stephen-cox (approved):
I've just tested this locally and it works fine happy for this to merged as.

Issue #270, opened 2023-07-17
Drupal 10 test failure


stephen-cox:
```
Drupal\Tests\localgov_alert_banner\FunctionalJavascript\AlertBannerHideTest::testAlertBannerHide
WebDriver\Exception\ElementNotVisible: element not interactable
  (Session info: chrome=77.0.3865.120)
  (Driver info: chromedriver=77.0.3865.40 (f484704e052e0b556f8030b65b953dce96503217-refs/branch-heads/3865@{#442}),platform=Linux 5.19.0-46-generic x86_64)

/app/vendor/instaclick/php-webdriver/lib/WebDriver/Exception.php:198
/app/vendor/instaclick/php-webdriver/lib/WebDriver/AbstractWebDriver.php:167
/app/vendor/instaclick/php-webdriver/lib/WebDriver/AbstractWebDriver.php:234
/app/vendor/instaclick/php-webdriver/lib/WebDriver/Container.php:241
/app/vendor/behat/mink-selenium2-driver/src/Selenium2Driver.php:805
/app/vendor/behat/mink-selenium2-driver/src/Selenium2Driver.php:791
/app/web/core/tests/Drupal/FunctionalJavascriptTests/DrupalSelenium2Driver.php:120
/app/web/core/tests/Drupal/FunctionalJavascriptTests/DrupalSelenium2Driver.php:200
/app/web/core/tests/Drupal/FunctionalJavascriptTests/DrupalSelenium2Driver.php:131
/app/vendor/behat/mink/src/Element/NodeElement.php:153
/app/web/modules/contrib/localgov_alert_banner/tests/src/FunctionalJavascript/AlertBannerHideTest.php:55
/app/vendor/phpunit/phpunit/src/Framework/TestResult.php:728
```

andybroomfield:
Is in `tests/src/FunctionalJavascript/AlertBannerHideTest.php`
````
    // Find and click hide link.
    $page = $this->getSession()->getPage();
    $button = $page->findButton('Hide');
    $this->assertNotEmpty($button);
    $button->click();
````

andybroomfield:
Could this be the old version of Chrome not working with Drupal 10?
(Chromedriver 77)

PR #272, opened 2023-07-25
Release 1.6.1

stephen-cox:


andybroomfield:
Review required? I think this change should work, though think it might be the older version of Chromedriver in the Drupal container? Is there a way to get screenshots of what chrome in the test suite is seeing?

stephen-cox:
I'm going to release this and then look at the test on #270

PR #273, opened 2023-07-31
Hide moderation controls when present on confirmation form page

### LINKED ISSUE #236: Content moderation form on manage display breaks the confirmation page
If its placed, you can't actually set the banner live on the confirmation screen.

_Originally posted by @andybroomfield in https://github.com/localgovdrupal/localgov_alert_banner/issues/174#issuecomment-1426916032_

andybroomfield:
Fix #236

Adds a preprocess implementation to the alert banner to hide the moderation controls if they have been displayed when on the confirmation page (status form).

Note: This only hides on the confirmation form page eg.
/admin/content/alert-banner/18/status

If the form is included in another way it will still be present.

This way, moderation controls can be placed on an alert banner manage display if a site builder considers that desirable, without interfering with the confirmation form to put an alert banner live.

### DIFF
--- a/localgov_alert_banner.module
+++ b/localgov_alert_banner.module
@@ -40,6 +40,13 @@ function localgov_alert_banner_preprocess(&$variables) {
     // Token as attribute.
     $variables['attributes']['data-dismiss-alert-token'] = $token;
 
+    // Remove the content moderation form if it is present.
+    // Do this only on the confirmation form page.
+    $route_match = \Drupal::routeMatch();
+    $route_name = $route_match->getRouteName();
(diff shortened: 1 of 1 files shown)


andybroomfield:
Test by adding a moderation control to the manage display and verify they are not present on the confirmation page, but can still be present under view.

stephen-cox (approved):
Approved. Removes the moderation control from alert banner previews on the set live form.

andybroomfield:
Thanks @stephen-cox Sorry I missed out on merge monday but I'll merge this now.

PR #274, opened 2023-08-04
Add enforced dependency on workflow

### LINKED ISSUE #213: The Localgov_alert_banners workflow should be removed when this module is uninstalled
Otherwise uninstall / reinstall fails.

Add the module as an enforced dependency in the `workflows.workflow.localgov_alert_banners.yml` file under dependencies.

````yaml
  enforced:
    module:
      - localgov_alert_banner
 ````

CATCHPOLEBHC:
fix #213

Uninstalling and reinstalling alert banner previously produced an error.  Now it uninstalls correctly.


andybroomfield:
From a BHCC dev so I think we need to have cross organisation review. Though I have checked and it fullfills the functionality (the workflow is removed with the module).

stephen-cox (approved):
Looks fine

PR #280, opened 2023-08-30
Add the is_front template flag using path.matcher service

### LINKED ISSUE #279: Homepage classes on alert banner template no longer showing
There should be extra classes added to the alert banner template `localgov-alert-banner--homepage` Should a theme wish to make alert banners on the homepage appear different (usually larger text). This is provided by a default Drupal twig variable `is_front`.

It looks like there has been some change in how the `is_front` twig variable is set, as it is not being sent be default.
This can be fixed by adding the following to the entity / preprocess
````php
try {
  $variables['is_front'] = \Drupal::service('path.matcher')->isFrontPage();
}
catch (Exception $e) {
  $variables['is_front'] = FALSE;
}
````
See 
https://drupal.stackexchange.com/a/204197
https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Path%21PathMatcher.php/function/PathMatcher%3A%3AisFrontPage/9

andybroomfield:
Fix #279

Same method as page manager.
This adds the is_front template flag to localgov alert banner for testing with twig.
Will restore the default homepage class in the default template.

### DIFF
--- a/localgov_alert_banner.module
+++ b/localgov_alert_banner.module
@@ -47,6 +47,14 @@ function localgov_alert_banner_preprocess(&$variables) {
     if ($route_name == 'entity.localgov_alert_banner.status_form') {
       unset($variables['content']['content_moderation_control']);
     }
+
+    // Set is_front variable.
+    try {
+      $variables['is_front'] = \Drupal::service('path.matcher')->isFrontPage();
+    }
+    catch (Exception $e) {
+      $variables['is_front'] = FALSE;
+    }
   }
 }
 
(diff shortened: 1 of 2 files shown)


markconroy (approved):
Looks good, let's merge.

andybroomfield:
I think I need to add a test? But then I will merge this.

andybroomfield:
Ready for re-review now the test is added.

finnlewis (approved):
Looks good to me, thanks @andybroomfield

PR #281, opened 2023-09-06
Add Github action on demand support to test

andybroomfield:
Adds the key `workflow_dispatch:` to the on section.


finnlewis:
That looks cool, once merged does this mean we can trigger the workflow through the Github Actions page?

andybroomfield:
Yes, its the same key we use at BHCC to allow us to run them on demand.

finnlewis (approved):
I totally approve!

PR #296, opened 2023-10-02
1.7.1

andybroomfield:
Quick path release for updating dev dependencies.


finnlewis (approved):
Good catch @andybroomfield !

PR #301, opened 2024-01-10
Fix running Paratest in Github Actions

stephen-cox:


andybroomfield (approved):
Tests appear to pass now which is good.

PR #306, opened 2024-02-05
[WIP] feat: use backend handling for cookie check

millnut:


andybroomfield:
What I currently see, the banner reappears when hidden.

https://github.com/localgovdrupal/localgov_alert_banner/assets/1467480/36642db9-269c-478b-ab43-7321e54541f2

This is with the new template addition in localgov_base and with internal page cache turned on.
Since anon users always get the same rendered page, the banner is always shown. Clearing the cache and refreshing does hide it, but it hides it for all anon users.

andybroomfield:
This will be [Drupals Internal page cache](https://www.drupal.org/docs/administering-a-drupal-site/internal-page-cache) which does not vary the page by cookie. It's only possible to bypass this by throwing the [page cache kill switch ](http://drupalplanet.com/blog/programmatically-disabling-page-caching-drupal-9-when-say-no-cache)with `\Drupal::service('page_cache_kill_switch')->trigger();` and any page with an alert banner would not be able to benefit from caching the rendered html to anon users. Given that the alert banner is quite often used by every page, I think this would be an hard performance hit, especially on smaller councils on non enterprise hosting, and even BHCC this would be a performance hit.

millnut:
Agreed @andybroomfield I spent some time looking into this more and trying different ways to invalidate and always hit the Drupal Internal Page Cache module for anonymous users. Some places recommend disabling this but I don't recommend this due to the negative performance impact you mentioned.

Thanks for testing this @andybroomfield I'll close/delete this PR later today and look into the client-side changes more.

PR #309, opened 2024-02-27
Update github ci to test php 8.1 and 8.2, just on Drupal 10.

### LINKED ISSUE #308: Test against PHP 8.2
Just noticed that we are testing on PHP 8.1 and 8.2 on localgov_project, but only PHP 8.1 here.

https://github.com/localgovdrupal/localgov_project/blob/3.x/.github/workflows/test.yml#L36

Maybe we should update to be consistent. 

@andybroomfield do you want to keep testing against Drupal 9  or happy to switch to just D10?

finnlewis:
Fixes #308

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -22,13 +22,13 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        include:
-          - localgov-version: '2.x'
-            drupal-version: '~9.4'
-            php-version: '8.1'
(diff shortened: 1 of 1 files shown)


andybroomfield (approved):
Thanks @finnlewis 
Unless there are specific objections I think its fine to drop 9 in the test, though I'll hold of on release unless there is a bug fix as I don't think we should Drop Drupal 9 support from module till next minor release.

PR #316, opened 2024-04-26
Fix: Group alert banner query alteration

Adnan-cds:
Covers special case where there are no alert banners present in any of the microsites.


Adnan-cds:
I have put some Band-aid around the code responsible for the fatal error.

stephen-cox (approved):
Works for me. I can now enable the `group_alert_banner` module and alert banners work as expected.

PR #319, opened 2024-04-30
Fix Translation for alert banner not displaying

### LINKED ISSUE #318: Translation for alert banner not displaying (Originally on drupal.org)
Original issue
https://www.drupal.org/project/localgov/issues/3427615

> Problem/Motivation
> Translations of alert banners are not being displayed.
> 
> Steps to reproduce
> 1. Create a new alert banner in English and publish it.
> 2. Translate it to a different language.
> 3. Specify in which pages the banner should be displayed.
> 4. Publish it.
> 
> Expected result: The original banner will be displayed in the English pages and so will be the translation in the language specific pages.
> Actual result: Only the original banner is displayed.
> 
> This happens because AlertBannerEntity::isVisible() is considering the conditions of the original banner, not the translation's.

[adinancenci](https://www.drupal.org/u/adinancenci) Also provided a [patch to test.](https://www.drupal.org/project/localgov/issues/3427615#comment-15487385)

andybroomfield:
Fix #318 

Alters the alert banner block to check for a translation, and add a translation test.

Originally authoured by [adinancenci](https://www.drupal.org/u/adinancenci).
Original patch in comment 3.
  https://www.drupal.org/project/localgov/issues/3427615#comment-15487385
(Patch originallty placed the logic in the entity, moved to block to fix issue with visibility checking).

### DIFF
--- a/src/Plugin/Block/AlertBannerBlock.php
+++ b/src/Plugin/Block/AlertBannerBlock.php
@@ -6,6 +6,7 @@ use Drupal\Core\Block\BlockBase;
 use Drupal\Core\Cache\Cache;
 use Drupal\Core\Entity\EntityTypeManagerInterface;
 use Drupal\Core\Form\FormStateInterface;
+use Drupal\Core\Language\LanguageManagerInterface;
 use Drupal\Core\Plugin\ContainerFactoryPluginInterface;
 use Drupal\Core\Session\AccountProxyInterface;
 use Drupal\field\Entity\FieldStorageConfig;
@@ -37,6 +38,13 @@ class AlertBannerBlock extends BlockBase implements ContainerFactoryPluginInterf
   protected $currentUser;
 
   /**
+   * The language manager service.
+   *
+   * @var \Drupal\Core\Language\LanguageManagerInterface
+   */
+  protected $languageManager;
+
+  /**
    * Constructs a new AlertBannerBlock.
    *
    * @param array $configuration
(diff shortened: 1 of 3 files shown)


src/Plugin/Block/AlertBannerBlock.php
```diff
@@ -192,8 +204,15 @@ protected function getCurrentAlertBanners() {
 
     // Load alert banners and add all.
     // Visibility check happens in build, so we get cache contexts on all.
+    $langcode = $this->languageManager->getCurrentLanguage()->getId();
     foreach ($published_alert_banners as $alert_banner_id) {
       $alert_banner = $this->entityTypeManager->getStorage('localgov_alert_banner')->load($alert_banner_id);
+
+      // Get alert banner translation if it exists.
+      if ($alert_banner->hasTranslation($langcode)) {
+        $alert_banner = $alert_banner->getTranslation($langcode);
```
ekes: Shouldn't this be EntityRepository::getTranslationFromContext()

and I assume this has knock on with the fact that this is called from the constructor (and probably shouldn't be).

andybroomfield: Yeah I was looking into that, I think we might need to get #321 merged first to check this (remove call from constructer).

That said, the alert banner loaded is the correct translation until the condition fields display paths are set to something different for the translation, so I assume there is already some translation context aware loading going on.

I added a test case for the scenarios I was seeing the break on and the above method does work. Is there a particular reason to favour `EntityRepository::getTranslationFromContext()` as opposed to `$alert_banner->getTranslation($langcode)` method?

ekes: > I added a test case for the scenarios I was seeing the break on and the above method does work. Is there a particular reason to favour EntityRepository::getTranslationFromContext() as opposed to $alert_banner->getTranslation($langcode) method?

Because the former is intended for getting translations to edit them, the later is intended for getting translations to display them in context should they exist. The later does the work of checking if there is a translation for you, checks cachability, and it uses the language fallback system (as configured for the particular site) to decide which language to use.

andybroomfield:
This is ready for re-review as I'd like to get this fixed and close the ticket on drupal.org.

ekes (approved):
Looks good :+1:

PR #322, opened 2024-05-03
Fix deprecations, replace cookies with localstorage

millnut:
- Fix a deprecation on `spaceless` that was missed (see https://www.drupal.org/project/drupal/issues/3094850)
- Remove `full-page-alert-banner.es6.js`, Drupal doesn't use es6 for builds anymore
- Replace cookie functionality with LocalStorage
    - Cookie storage limit 4KB
    - LocalStorage limit 5MB
- Refactor in plain/vanilla javascript to remove jQuery dependency


ekes (approved):
https://www.drupal.org/node/3322720

andybroomfield:
Thanks @millnut. 
I'd pause on this for now as js-cookie includes a remote CDN loading dependency unless people know to include it.
Since that would be springing it on people in a patch release, I'd rather wait [until that is released](https://www.drupal.org/project/js_cookie/issues/3390616) before we merge this.

I still think we can rewrite the cookie logic to not require js-cookie if someone has time.

andybroomfield:
Is [this template](templates/localgov-alert-banner-content-add-list.html.twig) actully used anywhere? it does not appear in any other custom entitiy. Maybe thats why we havn't noticed it.

millnut:
Thanks for checking @andybroomfield I wasn't aware it used a remote CDN. If I have the time I'll rewrite in plain javascript and we can probably remove this requirement altogether

ekes:
Is this still draft?

Notice similar but different work on  #347

millnut:
Yep still draft I need to do some checks on the tests

andybroomfield:
@ekes @millnut  #347 is just a quick and simple one to remove the js-cookie dependancy so I can get Drupal 11 compatabiliy. It would still be good to get this rewritten in vanialla JS and move to local storage at some point. Hopefully the change above does not cause too much conflict.

PR #324, opened 2024-05-19
feat: use shared workflow

millnut:
This PR switches the module to use the shared workflow test file provided in the new workflow repo https://github.com/localgovdrupal/localgov_shared_workflows.

This will allow us to maintain a single workflow testing file which we can use across modules using the GitHub shared workflows functionality.

Once this has been reviewed I'll look at switching the other modules to this shared workflow.


stephen-cox (approved):
Nice work @millnut! This is a huge simplification of the workflow management.

There maybe issues with the shared workflow when a module needs to support different versions of Drupal or LGD, but it's like this won't be required until we start adding Drupal 11 support and don't effect this repo now, so no action required, just flagging it for later.

millnut:
Thank you @stephen-cox yep I've started to look at Drupal 11 and we have quite a few contrib modules that we are blocked on. We should be able to expand this workflow with optional inputs to cover those cases.

PR #325, opened 2024-05-21
Follow up performance improvements

andybroomfield:
Remove protected array for current banners and move the cache contexts
into th build method. Also amend the kernal test to account for #cache
inside the render array.

### DIFF
--- a/src/Plugin/Block/AlertBannerBlock.php
+++ b/src/Plugin/Block/AlertBannerBlock.php
@@ -37,16 +37,6 @@ class AlertBannerBlock extends BlockBase implements ContainerFactoryPluginInterf
   protected $currentUser;
 
   /**
-   * Current alert banners.
-   *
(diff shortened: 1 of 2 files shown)


andybroomfield:
@rupertj have managed to remove the need for storing current alert banners as a class property. Turns out it's not stored anyway between the build and getCacheContexts methods, so just applied it to the build render array.

rupertj:
Looks good, although it'd be better to use CacheableMetadata, as that'd add all the relevant metadata from the node, not just contexts. (I'm thinking cache tags should come across too).

andybroomfield:
@rupertj Yeah, I was looking at that, though we already have a getCacheTags and I think they should bubble. I can look at those in part 3 along with the manager service.

rupertj (approved):

rupertj:
I've just realised we're talking about a custom entity, and I don't think it implements the interface required to use CacheableMetadata right now, so doing it like this MR does is best for now anyway.

PR #328, opened 2024-05-31
Re-add the getCacheContexts method to block

### LINKED ISSUE #327: [Regression] Alert banners are no longer showing correctly when pages are cached
Looks like a regression introduced in #325 

The removal of the `getCacheContexts` method on the alert banner block is sometimes preventing banners from displaying. Looks like this is a partular issue where Varish is used, but could also be the internal page cache.

Becuase of #154 we don't add the `url.path` cache context unless the banner appears on a page, otherwise the cache table will fill up with every page. *But* this means that the render array that is cached will be the one without this context unless its on the right page. 

It looks like `getCacheContexts` does get called and the value stored in the cache table, wheras doing this in build does not. I think at least for now we should readd `getCacheContexts` method. It will still call `$this->getCurrentAlertBanners()` but I don't see that as a major issue unless there are lots and lots of live banners.
(The banner entity loading is cached, so its just the query I think is not. Note that trying to store the banner as a protected property won't work as the block gets reconstructed between then getCacheContexts and build method).

Long term we should move to rendering the alerts through big pipe so the banner block is just a placeholder.

andybroomfield:
Fix #327

Part revert #325
Restores the getCacheContexts method as that is what is called each time to determine the cache contexts each time the block is rendered.

### DIFF
--- a/src/Plugin/Block/AlertBannerBlock.php
+++ b/src/Plugin/Block/AlertBannerBlock.php
@@ -146,9 +146,7 @@ class AlertBannerBlock extends BlockBase implements ContainerFactoryPluginInterf
 
     // Render the alert banner.
     $build = [];
-    $contexts = [];
     foreach ($published_alert_banners as $alert_banner) {
-      $contexts = Cache::mergeContexts($contexts, $alert_banner->getCacheContexts());
 
       // Only add to the build if it is visible.
       // @see #154.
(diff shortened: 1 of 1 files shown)


rupertj (approved):

rupertj:
I've done some manual testing on this and while I couldn't recreate the bug locally, the code in this PR works as expected. (And page cache is on for me, for the record).

I did find another cache bug - if you exclude a path, then view that path, then remove the exclusion, the block doesn't show until the cache is cleared, but that happens with or without this change, so it doesn't need to block this.

PR #330, opened 2024-06-04
Updating weight of local task from 50 to 30 (localgov Issue 718)

MariosORION:
This PR is part of https://github.com/localgovdrupal/localgov/issues/718.


finnlewis (approved):
Tested on Gitpod and works with the other pull requests on https://github.com/localgovdrupal/localgov/issues/718

PR #332, opened 2024-06-17
fix: additional strict type fixes

millnut:
Fixes the following errors, reported in https://github.com/localgovdrupal/localgov_project/pull/153

```
 ------ ---------------------------------------------------------------------------------- 
  Line   modules/contrib/localgov_alert_banner/src/Access/AlertBannerEntityPageAccess.php  
 ------ ---------------------------------------------------------------------------------- 
  25     Only booleans are allowed in &&, mixed given on the right side.                   
 ------ ---------------------------------------------------------------------------------- 

 ------ ------------------------------------------------------------------------ 
  Line   modules/contrib/localgov_alert_banner/src/Entity/AlertBannerEntity.php  
 ------ ------------------------------------------------------------------------ 
  152    Only booleans are allowed in an if condition, mixed given.              
 ------ ------------------------------------------------------------------------ 

 ------ --------------------------------------------------------------------------------------------------- 
  Line   modules/contrib/localgov_alert_banner/src/Form/AlertBannerEntityRevisionRevertTranslationForm.php  
 ------ --------------------------------------------------------------------------------------------------- 
  93     Only booleans are allowed in ||, mixed given on the right side.                                    
 ------ --------------------------------------------------------------------------------------------------- 
```

### DIFF
--- a/src/Entity/AlertBannerEntity.php
+++ b/src/Entity/AlertBannerEntity.php
@@ -149,7 +149,8 @@ class AlertBannerEntity extends EditorialContentEntityBase implements AlertBanne
     }
 
     // Regenerate a JS token for the updated alert banner.
-    if ($this->get('status')->value) {
+    $status = (bool) $this->get('status')->value;
+    if ($status) {
       $prefix = 'alert-' . $this->id() . '-';
       $hash = sha1(uniqid('', TRUE));
       $this->setToken($prefix . '-' . $hash);
--- a/src/Access/AlertBannerEntityPageAccess.php
+++ b/src/Access/AlertBannerEntityPageAccess.php
@@ -6,6 +6,7 @@ use Drupal\Core\Access\AccessResult;
 use Drupal\Core\Routing\Access\AccessInterface;
 use Drupal\Core\Routing\RouteMatchInterface;
 use Drupal\Core\Session\AccountInterface;
+use Drupal\localgov_alert_banner\Entity\AlertBannerEntity;
 
 /**
  * Checks access to alert banner entity pages.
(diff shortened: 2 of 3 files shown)


src/Entity/AlertBannerEntity.php
```diff
@@ -149,7 +149,8 @@ public function preSave(EntityStorageInterface $storage) {
     }
 
     // Regenerate a JS token for the updated alert banner.
-    if ($this->get('status')->value) {
+    $status = (bool) $this->get('status')->value;
```
millnut, line 152: We force a bool type here as previously it would either be `true/false` or `"1"/"0"`, this makes it less loose

andybroomfield:
Maybe we could target this to the  1.8.x branch as part of the refactor.

millnut:
> Maybe we could target this to the 1.8.x branch as part of the refactor.

Sorry @andybroomfield must have missed that what is the goal of the 1.8.x branch vs 1.x?

andybroomfield:
@millnut https://github.com/localgovdrupal/localgov_alert_banner/milestone/5

Its that the entity was generated with a very old version of Drupal console so need to redo quite a bit of it, have done some initial work in #333

millnut:
Great thank you @andybroomfield would you prefer another PR for 1.8.x? Slightly off topic of this PR but for Drupal 11 compatibility should I target 1.x or 1.8.x

andybroomfield:
@millnut I think we should focus Drupal 11 support and refactoring on 1.8.x so we can keep the 1.x branch for bugfixes.

andybroomfield (approved):
Approved, merging into 1.8.x refactor.

PR #333, opened 2024-06-17
Initial refactor of revision provider and access

### LINKED ISSUE #323: Do we need revisionOverview() in AlertBannerEntityController.php
Theres a deprecation being flagged for renderPlain to renderInIsolation in Drupal 10.3.

We might be able to just remove it though, no other custom entity has this.

andybroomfield:
Fix #323

Refactor the alert banner entitiy to use a more standard approach to revisions, using `\Drupal\Core\Entity\Routing\RevisionHtmlRouteProvider` instead of custom routes, and using standard revsion routes instead of custom versions in routing.yml.

This also groups all the non view crud permissions into the manage alert banner permission in alertBannerEntityAccessControlHandler.

This is the first part of a refactor for version 1.8.


andybroomfield:
Taking this out of draft. focusing this PR on just removing the history routes and using the standard Drupal provided ones.
Note: I'm targeting 1.8.x branch as this is part of a refactor. @localgovdrupal/maintainers are you happy with this approach as it means bug fixes can still be targeted against 1.x whilst this work is in progress, and I can do smaller PRs for specific issues. If so, do these indivudual ones still want reviews?

Adnan-cds:
I will leave it to you Andy :)

finnlewis:
Very happy to get the 1.8.x branch into a ready state for review. 

If we can get the tests running on merges INTO 1.8.x - that should help.

Issue #334, opened 2024-06-18
Drupal 11 support


andybroomfield:
Fixes here for Drupal 11 support

- [x] Replace jcookie with alternative
- [x] Fix entity so it no longer uses deprecated code
- [x] Fix strict types
- [x] Fix full screen alert banner js-cookie

millnut:
Hi @andybroomfield here are some more errors that need fixing For Drupal 11 which was reported by running upgrade status module on Drupal 10.3

```
web/modules/contrib/localgov_alert_banner/localgov_alert_banner.module | 146 | Construct empty() is not allowed. Use more strict comparison.
web/modules/contrib/localgov_alert_banner/modules/group_alert_banner/group_alert_banner.module | 106 | Construct empty() is not allowed. Use more strict comparison.
web/modules/contrib/localgov_alert_banner/modules/group_alert_banner/group_alert_banner.module | 107 | Construct empty() is not allowed. Use more strict comparison.
web/modules/contrib/localgov_alert_banner/src/Access/AlertBannerEntityPageAccess.php | 25 | Only booleans are allowed in &&, mixed given on the right side.
web/modules/contrib/localgov_alert_banner/src/AlertBannerEntityListBuilder.php | 20 | Implicit array creation is not allowed - variable $header does not exist.
web/modules/contrib/localgov_alert_banner/src/AlertBannerEntityListBuilder.php | 30 | Implicit array creation is not allowed - variable $row does not exist.
web/modules/contrib/localgov_alert_banner/src/AlertBannerEntityTypeListBuilder.php | 17 | Implicit array creation is not allowed - variable $header does not exist.
web/modules/contrib/localgov_alert_banner/src/AlertBannerEntityTypeListBuilder.php | 26 | Implicit array creation is not allowed - variable $row does not exist.
web/modules/contrib/localgov_alert_banner/src/Controller/AlertBannerEntityController.php | 93 | Implicit array creation is not allowed - variable $build does not exist.
web/modules/contrib/localgov_alert_banner/src/Controller/AlertBannerEntityController.php | 99 | Implicit array creation is not allowed - variable $build does not exist.
web/modules/contrib/localgov_alert_banner/src/Entity/AlertBannerEntity.php | 152 | Only booleans are allowed in an if condition, mixed given.
[...]

PR #336, opened 2024-06-23
Remove default_argument_skip_url from group manage alert banners view

andybroomfield:
Ref #334

Remove view default_argument_skip_url config key which is being removed.
See https://www.drupal.org/node/3382316


Adnan-cds (approved):
Thanks.

PR #346, opened 2024-07-05
Adds banner type as ARIA label to banner article elements

### LINKED ISSUE #345: Alert banners convey meaning using colour alone by types having styling reflecting their purpose
For example, _Death of a notable person_ uses a black background and white text, _Major_ a red background.

Setting a role and an accessible name (the alert banner type) on one the containing elements should fix.

danchamp:
Fixes #345:

- Sets banner type variable in template_preprocess
- Changes wrapper element to <section> and adds ARIA label and role

### DIFF
--- a/templates/localgov-alert-banner.html.twig
+++ b/templates/localgov-alert-banner.html.twig
@@ -32,7 +32,7 @@
   ]
 %}
 
-<section {{ attributes.addClass(classes) }} aria-labelledby="{{ 'type_of_alert_' ~ html_id }} {{ 'alert_' ~ html_id }}">
+<section {{ attributes.addClass(classes) }} aria-labelledby="{{ html_id ~ '-label' }}">
   {# Begin Wrapper #}
   <div class="localgov-alert-banner__wrapper">
     {# Begin Inner #}
@@ -40,12 +40,15 @@
       {# Begin Content #}
       <div class="localgov-alert-banner__content">
         {% if display_title %}
-          <h2 class="localgov-alert-banner__title" id="{{ 'alert_' ~ html_id }}">
+          <h2 class="localgov-alert-banner__title" id="{{ html_id ~ '-label' }}">
+            {# If a type of alert label is set add it for screen readers #}
+            {% if type_of_alert_label is not empty %}
+              <span class="visually-hidden">{{ type_of_alert_label }}: </span>
+            {% endif %}
             {{ content.title }}
           </h2>
         {% endif %}
 
-        <p class="visually-hidden" id="{{ 'type_of_alert_' ~ html_id }}">{{ type_of_alert_label|default('Announcement') }}</p>
         <div class="localgov-alert-banner__body">
           {{ content|without('title', 'link') }}
         </div>
--- a/templates/localgov-alert-banner.html.twig
+++ b/templates/localgov-alert-banner.html.twig
@@ -32,7 +32,7 @@
   ]
 %}
 
-<article {{ attributes.addClass(classes) }}>
+<article {{ attributes.addClass(classes) }} role="region" aria-label="{{ type_of_alert_label }}">
   {# Begin Wrapper #}
   <div class="localgov-alert-banner__wrapper">
     {# Begin Inner #}
--- a/templates/localgov-alert-banner.html.twig
+++ b/templates/localgov-alert-banner.html.twig
@@ -32,7 +32,7 @@
   ]
 %}
 
-<section {{ attributes.addClass(classes) }} role="region" aria-label="{{ type_of_alert_label }}">
+<section {{ attributes.addClass(classes) }} aria-label="{{ type_of_alert_label }}">
   {# Begin Wrapper #}
   <div class="localgov-alert-banner__wrapper">
     {# Begin Inner #}
--- a/templates/localgov-alert-banner.html.twig
+++ b/templates/localgov-alert-banner.html.twig
@@ -41,7 +41,7 @@
       <div class="localgov-alert-banner__content">
         <div id="{{ html_id ~ '-label' }}">
           {% if display_title %}
-            <h2 class="localgov-alert-banner__title" id="{{ html_id ~ '-label' }}">
+            <h2 class="localgov-alert-banner__title">
               {{ content.title }}
             </h2>
             {# If a type of alert label is set add it for screen readers #}
--- a/templates/localgov-alert-banner.html.twig
+++ b/templates/localgov-alert-banner.html.twig
@@ -39,15 +39,17 @@
     <div class="localgov-alert-banner__inner">
       {# Begin Content #}
       <div class="localgov-alert-banner__content">
-        {% if display_title %}
-          <h2 class="localgov-alert-banner__title" id="{{ html_id ~ '-label' }}">
-            {# If a type of alert label is set add it for screen readers #}
-            {% if type_of_alert_label is not empty %}
-              <span class="visually-hidden">{{ type_of_alert_label }}: </span>
-            {% endif %}
-            {{ content.title }}
-          </h2>
-        {% endif %}
+        <div id="{{ html_id ~ '-label' }}">
+          {% if display_title %}
+            <h2 class="localgov-alert-banner__title" id="{{ html_id ~ '-label' }}">
+              {{ content.title }}
+            </h2>
+          {% endif %}
+          {# If a type of alert label is set add it for screen readers #}
+          {% if type_of_alert_label is not empty %}
+            <p class="visually-hidden">{{ type_of_alert_label }}</p>
+          {% endif %}
+        </div>
 
         <div class="localgov-alert-banner__body">
           {{ content|without('title', 'link') }}
--- a/templates/localgov-alert-banner.html.twig
+++ b/templates/localgov-alert-banner.html.twig
@@ -44,10 +44,14 @@
             <h2 class="localgov-alert-banner__title" id="{{ html_id ~ '-label' }}">
               {{ content.title }}
             </h2>
-          {% endif %}
-          {# If a type of alert label is set add it for screen readers #}
-          {% if type_of_alert_label is not empty %}
+            {# If a type of alert label is set add it for screen readers #}
+            {% if type_of_alert_label is not empty %}
+              <p class="visually-hidden"> ({{ type_of_alert_label }})</p>
+            {% endif %}
+          {% elseif type_of_alert_label is not empty %}
             <p class="visually-hidden">{{ type_of_alert_label }}</p>
+          {% else %}
+            <p class="visually-hidden">{{ 'Announcement'|t }}</p>
           {% endif %}
         </div>
 
--- a/templates/localgov-alert-banner.html.twig
+++ b/templates/localgov-alert-banner.html.twig
@@ -31,8 +31,9 @@
     has_link ? 'localgov-alert-banner--has-link' : 'localgov-alert-banner--no-link'
   ]
 %}
+{% set alert_id = random() %}
 
-<section {{ attributes.addClass(classes) }} aria-label="{{ type_of_alert_label }}">
+<section {{ attributes.addClass(classes) }} aria-labelledby="{{ 'type_of_alert_' ~ alert_id }} {{ 'alert_' ~ alert_id }}">
   {# Begin Wrapper #}
   <div class="localgov-alert-banner__wrapper">
     {# Begin Inner #}
@@ -40,11 +41,12 @@
       {# Begin Content #}
       <div class="localgov-alert-banner__content">
         {% if display_title %}
-          <h2 class="localgov-alert-banner__title">
+          <h2 class="localgov-alert-banner__title" id="{{ 'alert_' ~ alert_id }}">
             {{ content.title }}
           </h2>
         {% endif %}
 
+        <p class="visually-hidden" id="{{ 'type_of_alert_' ~ alert_id }}">{{ type_of_alert_label|default('Announcement') }}</p>
         <div class="localgov-alert-banner__body">
           {{ content|without('title', 'link') }}
         </div>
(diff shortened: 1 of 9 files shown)


andybroomfield:
@danchamp You can ignore the test fail as it's fixed in 1.8.x. Depending on how soon you need this, you could change the base branch to 1.8.x instead on 1.x and make it part of the refactor work.

templates/localgov-alert-banner.html.twig
```diff
@@ -32,7 +32,7 @@
   ]
 %}
 
-<article {{ attributes.addClass(classes) }}>
+<article {{ attributes.addClass(classes) }} role="region" aria-label="{{ type_of_alert_label }}">
```
markconroy: If we intend for this to be a "region" should we remove `<article>` and replace it with `<section>`?

It seems strange to have an article element reassigned as a region.

I think MDN agrees with me: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/region_role#prefer_html

danchamp: @markconroy Thanks for the pointer, that would make sense and I'll update it.

danchamp: @andybroomfield Thanks, I'll move it to 1.8x, it's not urgent so can go with the refactor.

andybroomfield: The change to article (from div) was made in #123. I'm reading up on [MDN Region role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/region_role) which states that the `<section>` tag is preffered without a aria-role. TBH it's a little confusing as I thought it was best to avoid section if something more suitable was in places (maybe [note](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/note_role)? 

Is the reason to move away from article so you can apply an `aria-label` for screen readers? Would a span with screen reader text be more approriate, using the Drupal standard of `<span class="visually-hidden">label="{{ type_of_alert_label }}</span>` be more appropriate?

Should we make the block as a whole a region (aside?)

danchamp: @andybroomfield You're right, if we use `<section>` then the explicit role is redundant and should be removed. I've updated the template.

The benefit of using an element with an implicit or explicit role is that it makes it a navigational landmark and easier for AT users to be aware of. Example from the demo site with this is place and landmarks listed:

alert-banner-landmarks

`<article>` has no implicit role, and we'd be relying on users of AT to identify the type of alert by another route - headings perhaps, or by linear access of something like that hidden `<span>`.

There may be a better way to achieve the aim - a visible label with the alert type maybe? We're currently relying on sighted users to perceive and correctly interpret the colours for each type, so should maybe be more explicit to all users?

andybroomfield: What happens when you have more than one alert of the same type? do they repeat?

danchamp: Yes, each alert would be a separate landmark.

andybroomfield: @danchamp Just to check, this is correct and intentional (Annoucement present twice in landmarks navigation as there are two annoucement banners).
Screenshot 2024-07-10 at 11 54 54 AM

danchamp: @andybroomfield I think that's desirable, since each banner is self-contained, and hiding one shouldn't hide both landmarks. I have invited feedback from the resident accessibility experts!

waako: I feel like @andybroomfield is on the right track with outputting the type of alert as visually-hidden in the section itself.

However, in an ideal world, the section label should be unique where possible otherwise, as illustrated in last screenshots, you get no context beyond the type of alert.
Therefore, the optimal approach would probably be:  
- Generate unique ID for each alert
- output the type of alert: `<p id="{{ 'alert_type__' ~ alert_id }}">{{ type_of_alert_label }}</p>`  
- Inside `.localgov-alert-banner__content`, after the title but before body.   
  Added benefit is that if CSS broken (and therefore no colours) the alert type is available visually still.
- Add ID to `h2.localgov-alert-banner__title` using `alert_id`
- On `section` container, change `aria-label` to `aria-labelledby="{{ 'alert_type__' ~ alert_id }} {{ alert_title_id }}"`

Then at least if a title set, it's used to provide context in the landmark list rather than just repeated alert type.

Does that make sense, or overkill ?

waako: I just spotted the potentials issue list in Slack like
> Can we rely on there always being a type set for a banner?

Can we default to Announcement ? something like 
`<p id="{{ 'alert_type__' ~ alert_id }}">{{ type_of_alert_label|default('Announcement') }}</p>`  
not sure where we would put `|t` here unless define a variable for fallback text ?

danchamp: @waako Thanks for this, it all makes total sense. I've updated the template, leaving the translation issue for now:

alert-banner-unique-landmarks

Would some punctuation between the type and title in the ARIA label help here?

andybroomfield: @danchamp Since the alert banners are translatable and the fields your using are content, they should be the translated versions anyway (worth checking). I assume screenreaders arn't going to read the title out twice?

danchamp: @andybroomfield Yes to the content fields already being translatable, I think the only element that isn't covered is the hard-coded 'Announcement' fallback string if a type isn't set for some reason. It's an edge case though, and may not be worth the effort of adding it as a user-definable variable?

Screen readers shouldn't read the title twice, since the two places it's included are in different contexts - label for the landmark, and the actual content of the hidden para.

waako: TLDR; Keep type of alert label separate from alert title in html, and onward discussion needing in term of 1.4.1 Use of color

## Unique ID
I'd forgotten about Drupal's [`clean_unique_id` twig filter](https://www.drupal.org/docs/develop/theming-drupal/twig-in-drupal/filters-modifying-variables-in-twig-templates#s-clean-unique-id).
So shouldn't need the whole `html_id` logic, just `id="{{ alert_type|clean_unique_id }}"`.
But jumping ahead of myself probably best define an `alert_id`: `{% set alert_id = alert_type|clean_unique_id %}` then have flexibility to do `id="{{ alert_id ~ '-label'}}"`

## Type of alert label placement
I can fully understand why you would render the type of alert label inside the heading, as a prefix, however I worry that this will:
- Result in a non-equivalent experience between visual and non-visual users
- Be confusing to non-blind screen-reader users

Technically this is [not a failure of [2.5.3 Label in name](https://www.w3.org/WAI/WCAG21/Understanding/label-in-name#accessible-name-and-description-computation-specification), but does not feel right nevertheless.

My initial reasoning for outputting the `<p id="{{ alert_id ~ '--type' }}" class="visually-hidden">{{ type_of_alert_label }}</p>` after the title in source order, was that if screen-reader uses jumps to alert h2 directly, they still get the context of alert type when reading on from h2, but it's no hard rule.
Either way, adding punctuation like colon at end of type of alert label is perfectly reasonable here I'd say.

Of course having the type of alert label render as hidden text only available to certain assistive technologies does bring up the _(colour)_ elephant in the room

## Use of colour

The ideal solution to the general issue raised by #345 (1.4.1 Use of Color) would be to _visibly_ render the **type of alert label** on the page.
This would introduce a change to any existing sites and introduces design decisions that I don't feel in a position to answer, maybe @msayoung might have some thoughts on best approach to this, also if and whom would need buy-in from ?

Sorry for the essay, been cogitating this and it grew into a mammoth

andybroomfield: We can revert the last commit if we want to avoid the type inside the label.

Lets do the following:
- Keep the html_id inside preprocess as those who override the template will need access to it.
- The type of alert field is optional and may not always be present, especially on custom alert banners. It's presence cannot be replied upon and if it is missing it should not be assumed to be of a particular type so for now lets leave that off the label if it is not present. (It's canonically not of any type so assuming announcement is incorrect).
- I think displaying the type of alert is a product change and would have to go to product and content design teams, lets not force that on sites. Lets do the accessibility bug fix for now and take it to the product and content groups if they want to make it visible.
- Can we just use a single id for the label, even if it means a wrapper around the h2 and the type. `<div id="{{ html_id ~ '--label' }}"><h2>{{ title }}</h2><p class="visually-hidden">{{ type_of_alert_label }}</p></div>`

andybroomfield: Theres an additional problem in that the title is in fact entirly optional so we can't actually gurantee the element will be present.

andybroomfield: If navigation is by headers, would we actully need to change this to a region. I'm still feel that the landmark is the alert banner block as a whole, not the indivual banners (and there could be more than one).

andybroomfield: Demo of the problem (no title just gives region) Screenshot 2024-07-24 at 9 47 46 AM

Moving the type of alert outside the title does resolve, giving the original behaviour.
Screenshot 2024-07-24 at 9 48 38 AM

andybroomfield:
@willguv This is the alert banner issue with the missing type of alert label.
If we make a product change, it will be to include either the text of the type of alert or another non-colour way like an icon inside the banner.

templates/localgov-alert-banner.html.twig
waako: Thanks for the background on implementation, that's helpful

- I can see the logic in keeping `html_id` preprocess, reduce risk of forgetting to `set html_id` in the twig template override.
- The fact that title is optional was another reason I wanted the `type_of_alert_label` separate from `display_title`
- Alert type being optional as well means makes sense to provide fallback value, such as _Announcement_, then no risk of landmark just being "region".

> Can we just use a single id for the label, even if it means a wrapper around the h2 and the type. `<div id="{{ html_id ~ '--label' }}"><h2>{{ title }}</h2><p class="visually-hidden">{{ type_of_alert_label }}</p></div>`

Probably viable, although it would need to be outside the `{% if display_title %}` block.

> If navigation is by headers, would we actully need to change this to a region. I'm still feel that the landmark is the alert banner block as a whole, not the indivual banners (and there could be more than one).

The problem, as always, is the versatility of this module, if it was showing 4 "basic" announcements each with a title, then having the block container being `<section aria-label="Announcements">` would make sense and be less verbose for screen-reader users navigating by landmarks, as you've said they can navigate by headers within the landmark.
However, if there were 1 "critical" alert, 1 "minor" alert and 2 "announcements" then it seems important that the "critical" alert at least is identified separately.

Maybe a more helpful approach would be:
- Set block container div to `section` with `aria-label="Announcements"`
- Output the alert type (if set) in the alert banner as agreed above
- If Alert type: Emergency, then change `article` to `section` with `aria-labelledby="html_id"`

Would a minor alert or death of a notable person warrant being highlighted in landmarks as individual region ?

For the 1.4.1 Use of Color failure, we are agreed that's a bigger change and needs input on what visual mechanism is best or least impactful to implement.

andybroomfield: @waako See latest push, though I might now make an accessible label in preprocess instead and set an aria-label instead of a labelled-by.

waako: @andybroomfield `aria-labelledby` is preferred as it references HTML content on the page whereas `aria-label` is not translatable - not talking about `Drupal.t` here but via scripts or services (hosted, OS provided or third-party) etc.

I understand why you might want to use preprocess for the accessible label, to keep some logic out of twig template.
However, this brings us back to using 2 IDs for the aria-labelledby, it would simplify the output, and if no title then no ID, therefore ignored and just reads out `type_of_alert_label` value.

Would this work and appease keeping things in preprocess ?
- Define `alert_title_id` and `alert_type_label_id` in preprocess, so "if title" set `alert_title_id` to `html_id + '--title'` else `''` (empty string so ignored in aria-labelledby)
- Update the `type_of_alert_label` in preprocess to have default _Announcement_ value if type of alert not set

Then it simplifies the twig template logic

```twig
{% if display_title %}
    <h2 class="localgov-alert-banner__title" id="{{ alert_title_id }}">
      {{ content.title }}
    </h2> 
 {% endif %}
 <p class="visually-hidden" id="{{ alert_type_label_id }}">{{ type_of_alert_label }}</p>
 ```
No need for any wrapping div, and risk of nesting/logic confusion ?
p.s.: Spotted that currently the `id="{{ html_id ~ '-label' }}"` set both on `div` and `h2`

andybroomfield: How do we deal with a situation where there is no title and no label?

Screenshot 2024-07-24 at 1 06 15 PM

andybroomfield:
The test failure here is becuase the label can't be found as the defination is
`array(4) { ["00--announcement"]=> string(12) "Announcement" ["20--minor"]=> string(11) "Minor alert" ["50--major"]=> string(9) "Emergency" ["90--notable-person"]=> string(25) "Death of a notable person" }`

I think thats the test thats wrong though, although we should correct that as it is possible to set up a type of alert without a label.

andybroomfield:
I'm moving this back to draft whilst we continue work on it.

templates/localgov-alert-banner.html.twig
andybroomfield: Theres also the issue of translation that we need to check.... Not sure how thats going to work becuase these values are part of the field config.

andybroomfield:
Alt is #376 (just show it using manage display).
I'm concerned that the type of alerts are really internal names, and may not be suitable for public display.

andybroomfield:
My instinct is that we maybe should'nt set this here, and this is something a theme should handle, as we don't actully know where someone will place the block. I can see value in wrapping the whole alert banner block in a landmark and using headers for navigation... I do think we need to split this from the colour / label issue which we could do by just including it in manage display as per #376, see my concern about the type of alerts beiing internal names so might need renaming.

PR #347, opened 2024-07-08
Remove js-cookie dependancy

### LINKED ISSUE #271: JS Cookie library deprecated in Drupal 10.1, removed in Drupal 11 without replacement

andybroomfield:
Fix #271

Small refactor to remove the js-cookie dependancy by using native javascript cookie read and write methods.

### DIFF
--- a/js/alert_banner.js
+++ b/js/alert_banner.js
@@ -5,19 +5,28 @@
  * This is so if the alert changes, the banner is reshown.
  */
 
-(function($, cookies) {
+(function($) {
 
   'use strict';
 
   function setAlertBannerHideCookie(cookie_tokens, token) {
     cookie_tokens.push(token);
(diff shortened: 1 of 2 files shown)


andybroomfield:
@millnut Would you be ok with me merging this change whilst you are doing the localstorage thing, as this just removes the js-cookie dependancy.

millnut:
@andybroomfield yes that's fine as it will help us loose the js-cookie dependency

PR #355, opened 2024-07-26
Move the visible check to alert banner manager

### LINKED ISSUE #354: Move visible check into alert banner manager
Place this behind an option `check_visible`.

andybroomfield:
Fix #354

### DIFF
--- a/src/AlertBannerManager.php
+++ b/src/AlertBannerManager.php
@@ -40,6 +40,7 @@ class AlertBannerManager implements AlertBannerManagerInterface {
     // Set default options.
     $default_options = [
       'type' => [],
+      'check_visible' => FALSE,
     ];
     $options = array_merge($default_options, $options);
 
@@ -71,7 +72,7 @@ class AlertBannerManager implements AlertBannerManagerInterface {
       ->execute();
 
     // Load alert banners and add all.
-    // Visibility check happens in block build, so we get cache contexts on all.
+    // Visibility check happens separately, so we get cache contexts on all.
     if (!empty($published_alert_banners)) {
       foreach ($alert_banner_storage->loadMultiple($published_alert_banners) as $alert_banner) {
         $alert_banner = $this->entityRepository->getTranslationFromContext($alert_banner);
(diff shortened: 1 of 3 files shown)


ekes:
@millnut do you know what's happening here?

stephen-cox:
> do you know what's happening here?

Composer is failing to install the testing branch because the `LATEST_RELEASE` variable defaults to 1 when it can't determine the latest release through the API call.

See: https://github.com/localgovdrupal/localgov_shared_workflows/blob/75394bb8ccaae453e67062d0c6e5e66215c8a609/.github/workflows/test-module.yml#L56-L60

This bit of code is showing itself to be a bit flaky. We either need to make the API call more reliable or create a variable that allows us to override the default value. The latter might be easier.

millnut:
@stephen-cox I wonder whether we can use a github action step to get the latest tagged release? Would that work?

stephen-cox:
> I wonder whether we can use a github action step to get the latest tagged release?

The API call in the workflow is supposed to do this for the major release, but in this case, because the target branch contains a minor release number the command is failing. This looks to be due to the bash string operations not being able to handle this case. I think I have a fix for this and will create a PR in the shared workflows for it.

stephen-cox:
There's a fix for this here: https://github.com/localgovdrupal/localgov_shared_workflows/pull/4

If we're happy with this then we can merge and then rerun the tests (after resetting the cache variable) and tests should run.

andybroomfield:
This is update to include a test for the scenario we encounterd in #327 where the cache contexts of the block meant the banner did not show. Interestingly I found out that checking for visibility did not seem to matter, but that might be as we are not negating yet. However it is important that the containing block does have cache contexts and if this api is extended we might want to offer an api to get cache contexts.

andybroomfield:
Anything that @localgovdrupal/maintainers want to comment on, or can I merge this as this should be enough to get a 1.8.0 release next merge Tuesday.

andybroomfield:
Going ahead and merging this.

PR #360, opened 2024-08-05
sets full-page alert banner to center align, adds ::backdrop

### LINKED ISSUE #359: Full page alert banner does not cover the full page
Maybe I'm wrong in what I am expecting, but I thought the full page alert banner would cover the full screen and not let me see the rest of the page until I dismissed it.

What I am seeing in an alert banner placed inside a `dialog` element, which does take over the page but we can still see everything behind it.

I think we need - at minimum - to add a darker `::backdrop` to the alert-banner dialog so the rest of the screen is not so visible.

I think we should also adjust the `margin` so it's set to `auto` for this item, so it's placed in the middle of the screen vertically and horizontally.

Here's what I am currently seeing:

Screenshot 2024-08-05 at 14 55 35

---

### LINKED ISSUE #580: Full page alert banner does not cover the full page
Maybe I'm wrong in what I am expecting, but I thought the full page alert banner would cover the full screen and not let me see the rest of the page until I dismissed it.

What I am seeing in an alert banner placed inside a `dialog` element, which does take over the page but we can still see everything behind it.

I think we need - at minimum - to add a darker `::backdrop` to the alert-banner dialog so the rest of the screen is not so visible.

I think we should also adjust the `margin` so it's set to `auto` for this item, so it's placed in the middle of the screen vertically and horizontally.

Here's what I am currently seeing:

Screenshot 2024-08-05 at 14 55 35

---

markconroy:
Closes localgovdrupal/localgov_base#580 

## What does this change?

We have - in localgov_base - a CSS property for 

```
*,
*::before,
*::after {
  margin-top: 0;
}
```

This means in `localgov_base` our full page alert banner is placed horizontally centered, but vertically at the top since the default `margin: auto` for `dialog` is overridden.

It's worse for `localgov_scarfolk` since we have an override to say

```
.localgov-alert-banner {
  margin: var(--spacing);
}
```
This means the alert banner is at the top left of the page, with `1rem` of spacing around it. 

This PR adds the default `margin: auto` back to the alert banner if the banner is a dialog element.

It also adds a `::backdrop` property so we can have the rest of the page faded out behind the dialog.

## How to test

Create a full page alert banner, be delighted.

## Images
Before:

Screenshot 2024-08-05 at 14 55 35

After:

Screenshot 2024-08-05 at 15 37 03

---

### DIFF
--- a/css/localgov-alert-banner.css
+++ b/css/localgov-alert-banner.css
@@ -8,13 +8,21 @@
   background-color: #00856A;
 }
 
+dialog.localgov-alert-banner {
+  margin: auto;
+}
+
+dialog.localgov-alert-banner::backdrop {
+  background-color: rgba(0, 0, 0, 0.8);
+}
+
 .localgov-alert-banner,
 .localgov-alert-banner a {
   color: #fefefe;
 }
 
 /* Reset the title colour as some themes assign a colour to h2.
-   Also reset the body text colour as Olivero assigns a colour 
+   Also reset the body text colour as Olivero assigns a colour
    to .text-content. */
 .localgov-alert-banner .localgov-alert-banner__title,
 .localgov-alert-banner .text-content {


andybroomfield:
@markconroy Do you want to target this as a bugfix release to 1.7 or as part of the 1.8 refactor?

markconroy:
I don't mind which really. It will affect both versions so probably worth committing to one version and cherry picking to the other.

css/localgov-alert-banner.css
```diff
@@ -8,13 +8,21 @@
   background-color: #00856A;
 }
 
+dialog.localgov-alert-banner {
+  margin: auto;
+}
+
+dialog.localgov-alert-banner::backdrop {
+  background-color: rgba(0, 0, 0, 0.8);
+}
```
andybroomfield, line 17: I think this should go in full screen alert banner module, not the main stylesheet.

/modules/localgov_alert_banner_full_page/css/full-page-alert-banner-layout.css or full-page-alert-banner-theme.css.

andybroomfield:
Seems fine in Olleivo though, so @markconroy are we sure we don't want to place this in localgov_base instead.
Note: one of the refactor plans is to make CSS optional, so it could be that localgov_base disables the alert banner styles and provides it's own anyway?

andybroomfield:
Noting the PHPStan fail is fixed in the 1.8.x branch.

markconroy:
Good idea, let's move this to `localgov_base`

PR #362, opened 2024-08-05
Fix the contrast on the type of alert text on full page alert banners

### LINKED ISSUE #361: Full screen alert banner has low contrast text
Example

Screenshot 2024-08-05 at 9 18 59 PM

This is on minor alert, could be others. Minor alert should use black text.

andybroomfield:
Fix #361

## What does this change?

Use color: inherit instead of overriding with a value

## How to test

Add a full page alert banner on each alert status and verify it has high contrast text (white on most banners, black on minor alert).
__Do this with theme set to Olivero or simmilar instead of localgov_base__

Text on full page alerts is no longer low contrast.

Theme might be being overridden.

## Images

Screenshot 2024-08-05 at 9 56 43 PM
Screenshot 2024-08-05 at 9 57 06 PM

## Accessibility

Only testing contrast as that is the change in this PR.

-   [ ] [Tested with screen reader](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#screen-reader)
-   [ ] [Navigable with keyboard](https://github.com/guardian/accessibility/blob/main/people-and-technology/02-physical.md#keyboard)
-   [x] [Colour contrast passed](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#contrast)
-   [ ] [The change doesn't use only colour to convey meaning](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#use-of-colour)

### DIFF
--- a/modules/localgov_alert_banner_full_page/css/full-page-alert-banner-layout.css
+++ b/modules/localgov_alert_banner_full_page/css/full-page-alert-banner-layout.css
@@ -22,9 +22,7 @@ dialog + .backdrop {
 	background: rgba(0, 0, 0, 1) !important;
 }
 
-.localgov-alert-banner-full h1,
-.localgov-alert-banner-full p,
-.localgov-alert-banner-full a {
+.localgov-alert-banner-full h1 {
 	color: inherit;
 }
 
--- a/modules/localgov_alert_banner_full_page/css/full-page-alert-banner-layout.css
+++ b/modules/localgov_alert_banner_full_page/css/full-page-alert-banner-layout.css
@@ -25,7 +25,7 @@ dialog + .backdrop {
 .localgov-alert-banner-full h1,
 .localgov-alert-banner-full p,
 .localgov-alert-banner-full a {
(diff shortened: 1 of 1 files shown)


andybroomfield:
Also happy to make this a merge on 1.x if we want to get this fix out sooner in the 1.7.x series.

markconroy (approved):

Adnan-cds:
I like the alert banners screenshots above :D

Looks okay to me.  I have taken a screenshot of the browser's Inspector tool: 
screenshot-full-page-alert-banner-style-rules This is from a stock LocalGov Drupal site.  I am assuming this is the expected outcome.

andybroomfield:
@Adnan-cds I think so yes, as it should be getting the link colour from the main banner stylesheet, so we should be able to remove a from the override.

andybroomfield:
Have pushed latest which removes the a and p tag from being set with inherit as it's only the h1 that needs it set.

PR #364, opened 2024-08-11
fix/1.x 363 emergency publisher perms

### LINKED ISSUE #363: Additional permissions for emergency publisher role
Currently if a user only has the emergency publisher role, they can't publish any banners as they can't reach the menu item.

We need to add 
- 'access toolbar'
- 'view the administration theme'

possibly
- 'access content overview' (to get the menu item in /admin/content

andybroomfield:
- **Add access toolbar, view administation theme perms to emergeny publisher role**
- **Add access content overview permission**

Fix #363

## What does this change?

Adds the following permisions to the emergency publisher role

- 'access content overview'
- 'access toolbar'
- 'view the administration theme'

This is to allow users with just the emergency publisher role to access the manage alert banners screen from the toolbar menu.

## How to test

Create a user with just the emergency publisher role and check they can navigate to content -> alert banners in the top menu.

Users with this role can now access the alert banners screen on default setup.

Gives access to access content overview, which they should'nt really need. Risk is small since typically those with this role would have access to that anyway. However since this PR is about those with just this role, it does give rise to asking if a seperate menu item should now be provided.

## Images

n/a

## Accessibility

n/a

### DIFF
--- a/config/install/user.role.emergency_publisher.yml
+++ b/config/install/user.role.emergency_publisher.yml
@@ -11,7 +11,6 @@ label: 'Emergency publisher'
 weight: -2
 is_admin: null
 permissions:
-  - 'access administration pages'
   - 'access content overview'
   - 'access localgov alert banner listing page'
   - 'access toolbar'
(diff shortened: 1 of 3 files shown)


andybroomfield:
So this was due to changes to emergency publisher installing permissions that are not defined during the test runs.

finnlewis (approved):
Tested locally and works a charm, thanks @andybroomfield !

PR #366, opened 2024-08-18
[1.8.x] fix: add ignore next line

millnut:
## What does this change?

This tells PHPStan to ignore the use of `sha1` and `uniqid` as we are only using for hashing


finnlewis (approved):
Nice!

PR #369, opened 2024-08-29
deprecate block contructor old signiture (current user and entity repository)

andybroomfield:
## What does this change?

Might not be neseccary, but if custom modules are extending the alert banner block, this will mark the old class signiture as deprecated. See https://github.com/localgovdrupal/localgov_alert_banner/wiki/Change-to-alert-banner-block-signature

## How to test

Extend the block with the old signature and deprecation messages show up.

Any custom modules get updated to the new signature

This might not be the right way of handling this, some people may not update to the new signature.

## Images

Screenshot 2024-08-29 at 11 14 11 pm

## Accessibility

n/a

### DIFF
--- a/src/Plugin/Block/AlertBannerBlock.php
+++ b/src/Plugin/Block/AlertBannerBlock.php
@@ -4,9 +4,11 @@ namespace Drupal\localgov_alert_banner\Plugin\Block;
 
 use Drupal\Core\Block\BlockBase;
 use Drupal\Core\Cache\Cache;
+use Drupal\Core\Entity\EntityRepositoryInterface;
 use Drupal\Core\Entity\EntityTypeManagerInterface;
 use Drupal\Core\Form\FormStateInterface;
 use Drupal\Core\Plugin\ContainerFactoryPluginInterface;
+use Drupal\Core\Session\AccountProxyInterface;
 use Drupal\localgov_alert_banner\AlertBannerManagerInterface;
 use Symfony\Component\DependencyInjection\ContainerInterface;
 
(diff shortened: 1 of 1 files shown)


millnut:
This looks good to me @andybroomfield my only concern is that paratest hides deprecations but this will be visible elsewhere to the user so it should be ok. 

On paratest specific I wonder in this issue https://github.com/localgovdrupal/localgov_shared_workflows/issues/9 we go with Option 1 and switch shared workflows to use phpunit for modules only that way we have better visibility of deprecations.

andybroomfield:
Thanks @millnut. This isn't really intended to flag to ourselves as I don't think we extend the block, but I do know several councils / agencies have made custom alter banner blocks where they add some persistent alerts and extra information, so it's intended to indicate the new __construct signature and provide a graceful way of informing them and adding the alert banner manager, hopefully this is the right way to do that.

src/Plugin/Block/AlertBannerBlock.php
```diff
@@ -28,20 +30,6 @@ class AlertBannerBlock extends BlockBase implements ContainerFactoryPluginInterf
    */
   protected $entityTypeManager;
 
-  /**
```
ekes, line 31: Removing these already, when only deprecated, is also to highlight the change? (and I assume they aren't used in these implementations any more).

andybroomfield: There not actually set anymore by the alert banner, though I could re-add them if needed (the functionality has been taken over by the alert banner manager).

andybroomfield:
This is the guide I'm reading https://www.drupal.org/about/core/policies/core-change-policies/how-to-deprecate

src/Plugin/Block/AlertBannerBlock.php
ekes: :+1:

ekes (approved):

src/Plugin/Block/AlertBannerBlock.php
andybroomfield: So I need to add this trait https://www.drupal.org/about/core/policies/core-change-policies/how-to-deprecate#how-injected-service

PR #370, opened 2024-08-30
Fix todos in admin view functional test

alansaun92:
I noticed a couple of todos mentioned in the functional test AdminViewTest.

So I have moved the creation of the admin user into the setup function and added in a kernel test to confirm that the admin url that we are expecting is the one set on the manage alert banners view.

### DIFF
--- a/tests/src/Functional/AdminViewTest.php
+++ b/tests/src/Functional/AdminViewTest.php
@@ -47,8 +47,6 @@ class AdminViewTest extends BrowserTestBase {
    * Test load the alert banner admin view.
    */
   public function testLoadAdminView() {
-    // Create an admin user -- @todo move to set up.
-    $adminUser = $this->createUser([], 'admintestuser', TRUE);
     $this->drupalLogin($this->adminUser);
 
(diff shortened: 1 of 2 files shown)


ekes (commented):
Does the AdminViewUrlTest::testAdminViewUrl() need to be a whole new test, rather than another method on AdminViewTest?

andybroomfield:
@ekes, though the new admin view url is a kernel test so should be quicker (the AdminViewTest is a functional test). If we wanted to keep it there we could add the code to the method?

andybroomfield (approved):
I'm happy with this. We could move this back into the functional AdminViewTest but technically the admin alert banner view is optional, as there is a default alert banner entity collection provided, so this is testing that the alert banner view is set up correctly to the desired URL (overriding the entity collection).

ekes:
> @ekes, though the new admin view url is a kernel test so should be quicker (the AdminViewTest is a functional test). If we wanted to keep it there we could add the code to the method?

I'd overlooked it was a Kernel test.

All good.

PR #376, opened 2024-09-09
Show type of alert in the alert banner

### LINKED ISSUE #345: Alert banners convey meaning using colour alone by types having styling reflecting their purpose

andybroomfield:
Fix #345 

- **Support showing type of alert in the banner template**
- **Show the type of alert in the banner by default**

## What does this change?

Provides a default space in the template to show the output of type of alert. This uses a value from manage display, so it will only show if configured to do so. This will be the default in new installs only.

## How to test

1. To fresh install
2. Choose a non localgov_base derived theme (localgov_base overrides the template).
3. Create an alert banner of each type
4. View alert banner, which should now show the type of alert.

Fixes accesbility issue identified in #345 , allbeit in this case without the role as its just plain text.

This might not be the right solution, it won't effect current installs, it won't override existing themes.
However in this case it's more about providing an accessible default, as the issue identified in #345 is likley the themes responsibility?

## Images

Screenshot 2024-09-09 at 9 17 18 PM

## Accessibility

-   [x] [Tested with screen reader](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#screen-reader)
-   [x] [Navigable with keyboard](https://github.com/guardian/accessibility/blob/main/people-and-technology/02-physical.md#keyboard)
-   [x] [Colour contrast passed](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#contrast)
-   [x] [The change doesn't use only colour to convey meaning](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#use-of-colour)

### DIFF
--- a/config/install/core.entity_view_display.localgov_alert_banner.localgov_alert_banner.default.yml
+++ b/config/install/core.entity_view_display.localgov_alert_banner.localgov_alert_banner.default.yml
@@ -33,7 +33,7 @@ content:
     type: link
     region: content
   short_description:
-    weight: 1
+    weight: 2
     label: hidden
     settings: {  }
     third_party_settings: {  }
@@ -47,8 +47,14 @@ content:
     settings:
       link_to_entity: false
     third_party_settings: {  }
+  type_of_alert:
+    type: list_default
+    label: hidden
+    settings: {  }
+    third_party_settings: {  }
+    weight: 1
+    region: content
 hidden:
   content_moderation_control: true
-  type_of_alert: true
   uid: true
   visibility: true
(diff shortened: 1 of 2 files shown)


andybroomfield:
Are the names:
- Announcement
- Minor alert
- Emergency
- Death of a notable person
Suitable for public display? They where only intended for internal use.

PR #379, opened 2024-09-23
removes polyfills and es6 code

### LINKED ISSUE #378: Can we remove the polyfills and es6.js files from alert banner?

markconroy:
Closes #378 

## What does this change?

Removes polyfill files and ES6 specific files, so we can just use modern JS instead.

## How to test

Create some alert banners, make sure they still work

---


andybroomfield:
Updated base so this goes into 1.8.x.

The main js looks good (swapping var for const).
I'll let @Adnan-cds look at the full screen js as I think hes more familliar with it.

Adnan-cds (approved):
Works for me.  No errors or warnings in browser's Javascript console. Code changes also look good.  Coding standards failures are unrelated.  Thank you :)

andybroomfield:
The conflicts come from this branch originating from 1.x, but the target is 1.8.x since the refactor is changing the javascript. I'd rather we target into 1.8.x and push to get that released as it also removes the js-cookie dependancy. It might be easier to open a new PR sourcing from 1.8.x so it applies to refactored js. Looks like it's mostly convertng var to const?

markconroy:
New PR created from `1.8.x` https://github.com/localgovdrupal/localgov_alert_banner/pull/381

PR #386, opened 2024-10-14
Feature/201 remove jquery

### LINKED ISSUE #201: Convert the alert banner dismiss script to vanilla Javascript
Currently [/js/alert_banner.js](https://github.com/localgovdrupal/localgov_alert_banner/blob/32b3ce8c28345673711c5d8278a043940c837ae8/js/alert_banner.js) uses jQuery, when really it should just use vanilla javascript.

markconroy:
Closes #201 

## What does this change?

1. Removes jQuery
2. Add Drupal behaviour
3. Uses once() to get list of alert banners
4. Uses once() to get list of alert banner close buttons
5. Uses a template literal to write the `document.cookie`.

## How to test
Check that all the alert banners work as they currently do.

---

### DIFF
--- a/js/alert_banner.js
+++ b/js/alert_banner.js
@@ -5,45 +5,72 @@
  * This is so if the alert changes, the banner is reshown.
  */
 
-(function($) {
+(function localgovAlertBannerScript(Drupal) {
+  Drupal.behaviors.localgovAlertBanners = {
+    attach(context) {
+      "use strict";
 
-  'use strict';
-
-  function setAlertBannerHideCookie(cookie_tokens, token) {
-    cookie_tokens.push(token);
-    const new_cookie = cookie_tokens.join('+');
-    // Set expiry 30 days.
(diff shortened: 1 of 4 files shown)


andybroomfield:
I figured this out by using stark as my theme on a Localgov Drupal install.
The alert banners where still hidden, and the javascript failed as `once` was not defined.
Added the correct dependencies `core/drupal` and `core/once` and removed `core/jquery` from the libraries.yml file and now the banner works. I assume the theme also includes those libraries which would explain why it worked.

This should (I think?) also resolve any issue brough about by this Drupal core change: [Asset ordering is now more strictly determined by library dependencies
](https://www.drupal.org/node/3473558)

andybroomfield:
@markconroy In terms of this PR. If it works I'm happy to approve and merge. Prob won't do a release till the new year as I won't be around to debug and most councils have now entered change freeze so won 't be doing any further deploys anyway. That then gives time to sort out the linting issues seperatly.

andybroomfield (approved):

PR #391, opened 2024-11-28
sets field labels to hidden

### LINKED ISSUE #374: Full page alert banner - field labels should not be set to visually hidden
The display of the fields labels on all the fields on a full page alert banner are set to Visually Hidden. 

See the Manage display configuration /admin/structure/alert-banner-types/localgov_full_page/edit/display 

This results in additional unnecessary information for assistive technology and unneeded markup. The semantic HTML conveys all the necessary information about what the fields are. 

Recommendation: the the field labels to Hidden.

markconroy:
Closes #374 

## What does this change?

Sets field labels for Full Page Alert Banner to 'Hidden'.

Do we need to run an update hook for existing sites for this item?

---


andybroomfield:
@Adnan-cds Could you look at this as I think Croydon is main council using full page alert banner. First glace this looks ok.

@markconroy I'd imagine we don't want to change anyones custom display mode though so unless we have a way to detect default configuration best to leave as is.

msayoung (approved):

finnlewis:
We still want to wait for @Adnan-cds to double check this doesn't break anything!

Adnan-cds (approved):
Works for me.

Adnan-cds:
Please note that the Image item in the banner is a Media entity reference.  This change removes the hidden label of this Media entity reference.  But the **Image** field within the Media entity still renders with a hidden label which we can't remove from this module unless we roll a new Entity display mode for the Media image entity.

PR #402, opened 2025-01-09
Fix alert banner cookie path

### LINKED ISSUE #401: Regression: Hide alert banner cookie is limited by path
## Steps to reproduce

1. Set up a global alert baner (should be on all pages).
2. Go to any page that isn't the home page
3. Hide the banner
4. Go to home page

## Expected result
Banner should be hidden.

## Actual result
Banner is visible

This can be checked by going to the page where the banner was hidden. The cookie has a path of the page instead of the hone page.

- [x] Write test to verify this
- [x] Write fix (should be just adding path to the cookie.

andybroomfield:
Fix #401

## What does this change?

Fix the regression wherby hiding an alert banner was hidden on a sub page, the banner would appear on the home page. This was becuase the cookie was not setting a path since the move away from js-cookie.

## How to test

1. Set up a global alert baner (should be on all pages).
2. Go to any page that isn't the home page
3. Hide the banner
4. Go to home page

Regression fixed.

N/a.

### DIFF
--- a/js/alert_banner.js
+++ b/js/alert_banner.js
@@ -12,7 +12,7 @@ function setAlertBannerHideCookie(cookieTokens, token) {
   const expiry = Date.now() + 30 * 24 * 60 * 60 * 1000;
   document.cookie = `hide-alert-banner-token=${newCookie}; expires=${new Date(
     expiry
-  ).toUTCString()}; SameSite=Lax;`;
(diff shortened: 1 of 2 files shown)


ekes (approved):
[x] Manually tested - also for use case not using frontpage.
 [x] Automated test makes sense

PR #404, opened 2025-01-19
Auto fix of linting issues

### LINKED ISSUE #399: Fix linting issues
The included JS and CSS need to be linted after #386 is merged.

andybroomfield:
Fix #399 

## What does this change?

Apply auto fixes to alert banner js files using
```
npx eslint@8 --ext .js --resolve-plugins-relative-to=./web/core --fix web/modules/contrib/localgov_alert_banner
npx eslint@8 --ext .yml --resolve-plugins-relative-to=./web/core --fix web/modules/contrib/localgov_alert_banner
npx stylelint --ignore-path web/core/.stylelintignore --config web/core/.stylelintrc.json --fix web/modules/contrib/localgov_alert_banner/**/*.css
```

## How to test

ESLint tests will now pass (Javascript has 3 warnings)

Tests pass and banner still works as expected.

Need to make sure we test for any breaks

### DIFF
--- a/config/install/user.role.emergency_publisher.yml
+++ b/config/install/user.role.emergency_publisher.yml
@@ -5,7 +5,7 @@ dependencies:
     module:
       - localgov_alert_banner
   config:
-   - workflows.workflow.localgov_alert_banners
+    - workflows.workflow.localgov_alert_banners
 id: emergency_publisher
 label: 'Emergency publisher'
 weight: -2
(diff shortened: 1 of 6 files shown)


ekes (approved):
Can't see anything other than changes of line order, line breaks, single quotes, and case. Passes tests.

Adnan-cds (approved):
Agree with Ekes.  Thanks for the fix :)

PR #410, opened 2025-03-10
Feature/408 remove version

### LINKED ISSUE #408: Remove VERSION from libraries.yml
The `version: VERSION` lines in libraries.yml create a version query string using the Drupal Core version number for the asset files. This causes the files to be cached in the browser and will only invalidate when the core version changes due to the dependency on the version.

```html
<script src="/modules/contrib/localgov_alert_banner/js/alert_banner.js?v=10.4.1"></script>
```

When using edge caching or CDN caching this makes it difficult to invalidate the cache if there are changes to these assets and it requires users to manually clear their browser cache to get updated asset files.

By removing the `VERSION` Drupal will generate a hash id to the file which can be invalidated and new hash IDs generated when the Drupal cache is cleared e.g. `drush cr`

The string changes on every update or full cache flush, forcing browsers to load a new copy of the files as the URL changed.

> See: web/core/lib/Drupal/Core/Asset/JsCollectionRenderer.php:62

Examples with no `VERSION` set showing changes based on cache clears;

```html
<script src="/modules/contrib/localgov_alert_banner/js/alert_banner.js?sqjnfy"></script>

<script src="/modules/contrib/localgov_alert_banner/js/alert_banner.js?sqjnm9"></script>
```

markconroy:
Closes #408


andybroomfield:
Hmm, looks like a lot of strange PR commits in this? Maybe this needs a rebase before merge.

markconroy:
@andybroomfield I'll close this, in favour of this https://github.com/localgovdrupal/localgov_alert_banner/pull/411

PR #416, opened 2025-04-14
Update LocalgovDrupal README.md

andybroomfield:
Replace with LocalGov Drupal

## What does this change?

Fix the title, so it matches what the brand is.

## How to test

Check the readme is correct.

No more OCD triggers for fellow maintainers.

n/a.

## Images

n/a

## Accessibility

n/a

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
-# Localgov Drupal Alert banner
+# LocalGov Drupal Alert banner
 
-Localgov Drupal Alert banner module, adds a global alert banner block and entity.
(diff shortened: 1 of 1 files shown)


ekes:
Does it not get capitalised, but still with the space? LocalGov Drupal

andybroomfield:
I think this might be redundant as Alert banner gets it correct.

andybroomfield:
Ahh no, it's capitalised by with a space like @ekes says.

finnlewis (approved):
Lovely!

PR #419, opened 2025-05-06
Remove vestiage entity settings form

### LINKED ISSUE #418: Does AlertBannerEntitySettings.php do anything
This is a form and routed via AlertBannerEntityHtmlRouteProvider
```php
if ($settings_form_route = $this->getSettingsFormRoute($entity_type)) {
      $collection->add("{$entity_type_id}.settings", $settings_form_route);
    }
```
Commit note is:
> Restore settingsForm route
> 
> Note: this may still get removed later in the refactor, but for now restore
> route admin/structure/alert-banner-types

Other than that is was added 5 years ago, presumably through Drupal code generator.

andybroomfield:
Fix #418

Currently this is (I believe) unused and just has placeholder methods, should be safe to remove.
(Does anyone know what it would actually do).

## What does this change?

Removes the entity settings form.

## How to test

Does the alert banner still work

Less code to maintain!

Maybe this actually does something important.


finnlewis (approved):
Tested locally and all seems to work as expected!

PR #426, opened 2025-09-08
fix: failing phpunit test due to changes in scheduled transitions module

millnut:
## What does this change?

Fixes #424 due to upstream changes to the scheduled transitions module (see https://www.drupal.org/project/scheduled_transitions/issues/3094322) and related test changes (https://git.drupalcode.org/project/scheduled_transitions/-/merge_requests/71/diffs#7ca217a4d73f18e3d48666e05a4867cf9f6e71b6)

PR #427, opened 2025-09-08
fix: supress phpstan errors on drupal.org

millnut:
## What does this change?

This will supress the PHPStan errors on Drupal.org due to localgov_microsites_group and it not being a composer dev dependency (https://git.drupalcode.org/project/localgov_alert_banner/-/jobs/6424731).

It is unlikely we'd want to have this module as a dev dependency due to the volume of other modules it would bring in.

**Workflow testing errors are not due to this PR and are fixed by https://github.com/localgovdrupal/localgov_alert_banner/pull/426**


andybroomfield (approved):
Looks good to me!

andybroomfield:
We will find out https://git.drupalcode.org/project/localgov_alert_banner/-/merge_requests/4/pipelines

PR #430, opened 2025-11-16
fix: add return types and strict types

millnut:
## What does this change?

This fixes the return types that we had in the PHPStan baseline file as well as introducting the strict types declaration

### DIFF
--- a/localgov_alert_banner.module
+++ b/localgov_alert_banner.module
@@ -5,6 +5,8 @@
  * Contains localgov_alert_banner.module.
  */
 
+declare(strict_types=1);
+
 use Drupal\Core\Cache\Cache;
 use Drupal\Core\Routing\RouteMatchInterface;
 use Drupal\user\RoleInterface;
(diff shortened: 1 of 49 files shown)


finnlewis (approved):
Looks good, alert banners still work fine locally and all the tests pass.

Happy to approve, but would like @andybroomfield's eyes on it too :)

andybroomfield (approved):
Seems, sensible, thanks @millnut 
I haven't tested this with the group alert banner, but I assume the same applies there.
I've got quite a bit queued up so this is maybe worth a release before starting new work.

localgov_alert_banner.module
```diff
@@ -22,13 +24,14 @@ function localgov_alert_banner_help($route_name, RouteMatchInterface $route_matc
       return $output;
 
     default:
+      return NULL;
```
ekes, line 27: Should this be `''`

andybroomfield: I'll take a look when I start converting these to OOP hooks. Can't see any side effect. I don't think most LGD modules include a Hook help so maybe we remove this?

